PR #24, opened 2023-05-20
fix(serverless): input key error

### LINKED ISSUE #23: bug: KeyError 'input'
I've noticed recently that this error occurs within the library and thought I'd open an issue. It seems that input sometimes does not exist in job, specifically in old jobs that have been in the queue for a while. Therefore an exception occurs on this line: 
https://github.com/runpod/runpod-python/blob/main/runpod/serverless/work_loop.py#L44 
```py
> if job["input"] is None:
	 log.error(f"Job {job['id']} has no input parameter provided. Skipping this job.")
	 continue
```
The reason I opened an issue before a pull request is because input should always exist, even with old jobs, so I would like your input on this issue.

Thanks,
Dan

dan-online:
Resolves #23

### DIFF
--- a/runpod/serverless/work_loop.py
+++ b/runpod/serverless/work_loop.py
@@ -41,7 +41,7 @@ async def start_worker(config):
                 log.info("No job available, waiting for the next one.")
                 continue
 
-            if job["input"] is None:
+            if "input" not in job or job["input"] is None:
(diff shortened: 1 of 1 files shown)


justinmerrell:
Functionality implemented with the following: https://github.com/runpod/runpod-python/commit/900ad0dcdb8033e97ab32c8239248b32f1684d85

PR #38, opened 2023-06-22
Handle 401 error for serverless calls

arsenyinfo:


justinmerrell:
Thank you for the PR!

PR #49, opened 2023-07-06
48 heartbeat refactor

### LINKED ISSUE #48: Heartbeat Refactor
Rename for consistency heartbeat.py > rp_ping.py 

Proposed changes:
- Move ping URL constructor completely out of worker_state and into rp_ping 

We can leave worker_state as the source where jobs currently in progress will be stored for now.

justinmerrell:
closes #48


runpod/serverless/modules/worker_state.py
```diff
@@ -40,30 +22,36 @@ def get_auth_header():
...
+class Jobs:
+    ''' Track the state of current jobs.'''
 
-def get_done_url():
-    '''
-    Constructs the done URL using the current job id.
-    '''
-    return JOB_DONE_URL_TEMPLATE.replace('$ID', CURRENT_JOB_ID)
+    _instance = None
+    jobs = []
```
Jorghi12: Looks good! One precondition for proper execution of this code region (during multi-job per worker scenario) is the assumption that JobTake will always return unique job_id values (e.g. it'll never return the same job_id more than once). Otherwise, it can lead to problems with add_job, remove_job, and the information sent via our heartbeat.

Violation of the pre-condition could result in adding the same job_id twice within a window. Calling remove_job once would delete both job_ids, and calling it the second time would cause a ValueError exception. Additionally, a race condition would be possible, where the list of jobs sent via ping can have the same job_id repeated twice.

If JobTake is guaranteed to always returns a job_id at most once to a single worker, then we can safely ignore above!

Jorghi12 (approved):
I briefly reviewed the heartbeat refactor, mentioning a scenario with the Jobs singleton and its assumptions on JobTake. Other than that, everything looks great!

runpod/serverless/modules/worker_state.py
```diff
@@ -40,30 +22,36 @@ def get_auth_header():
...
-def get_stream_url():
-    '''
-    Constructs the stream URL using the current job id.
-    '''
-    return JOB_STREAM_URL_TEMPLATE.replace('$ID', CURRENT_JOB_ID)
+    def add_job(self, job_id):
+        '''
+        Adds a job to the list of jobs.
+        '''
+        self.jobs.append(job_id)
```
justinmerrell: @Jorghi12 What if we do `self.jobs = set(self.jobs.append(job_id))` here to enforce id uniqueness?

justinmerrell: Additionally, we can check if a job is already in the list when it does get_job

Jorghi12: Interesting. Sure, we could initialize `self.jobs = set()` and do `self.jobs.add(job_id)` and then yeah checking if the job is already in the list when doing `get_job` works!

PR #58, opened 2023-07-13
fix fn check if it is a generator

### LINKED ISSUE #57: Generator support changes incorrect
**Describe the bug**
from the original PR I submitted, the code looked like https://github.com/runpod/runpod-python/blob/f19342c33dff2ce15ff93f85f92ea80427fb614a/runpod/serverless/work_loop.py#L63-L71

and now is https://github.com/runpod/runpod-python/blob/main/runpod/serverless/work_loop.py#L66-L72

Logically, this is incorrect, as you can't test the handler function is a GeneratorType, you have to actually call the function, and only then can you determine if it is a GeneratorType from the result of the function call.

**To Reproduce**

**Expected behavior**

**Screenshots**
Screenshot 2023-07-13 at 2 09 31 AM

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

winglian:
fixes #57 

`isinstance(config["handler"], types.GeneratorType)` doesn't work as expected in this case because it should be run against the job result. using inspect is the proper way to handle this after the previous refactor was applied

### DIFF
--- a/runpod/serverless/work_loop.py
+++ b/runpod/serverless/work_loop.py
@@ -2,7 +2,7 @@
 runpod | serverless | worker_loop.py
 Called to convert a container into a worker pod for the runpod serverless platform.
 """
-
+import inspect
 import os
 import sys
 import types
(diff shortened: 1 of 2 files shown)


winglian:
There is also another bug in the code where the streaming callback url is not correct and simply sending to the completed callback

runpod/serverless/work_loop.py
MMMartt, line 1: `job_result` in generation function branch will be set to `None`, which will raise an error in `local_test`(see L97). Maybe it should be fixed.

justinmerrell:
Partial fix in main, please rebase the PR.

runpod/serverless/work_loop.py
justinmerrell: @Jorghi12 I saw you also made some changes around this, did you cover the local_test condition?

justinmerrell:
Functionality in main, thank you for the PR!

PR #61, opened 2023-07-18
added documentation for Endpoint and create_pod

therealadityashankar:


runpod/endpoint/runner.py
```diff
@@ -1,7 +1,8 @@
 '''
 RunPod | Python | Endpoint Runner
 '''
-
+from __future__ import annotations
```
justinmerrell: Why do we need __future__? I would like to support some older versions of python, however, could we do this as a separate PR?

justinmerrell (changes requested):

runpod/endpoint/runner.py
therealadityashankar: Hey, `from __future__ import annotations` is supported in all python versions > python 3.7, I'm not sure if we're looking to support < python3.7, I believe we already don't and I'd be surprised if requests supports <python3.7 too

this basically makes it possible to add type hints for a class described after the type was mentioned

so like

this would not work

```python
class A:
     def spam() -> B:
           return B()

class B:
     pass
```

will not work, but

```python
from __future__ import annotations

class A:
     def spam() -> B:
           return B()

class B:
     pass
```

will work

therealadityashankar: yeah we only support python 3.7+ (https://pypi.org/project/requests/, needs python3.7+ and we have it as a dependency), so this will not change anything

justinmerrell: Right now, this repo only supports 3.10+ because of some other functionality. 

I would like to evaluate support for 3.8+ but in a separate PR that covers the whole repo. For this particular PR does your changes work without __future__ for 3.10?

therealadityashankar: no, my changes will not work in python 3.10 without the future import

justinmerrell: Maybe i am missing something, is that not what is in this file without __future__? https://github.com/runpod/runpod-python/blob/cf7dae6be8aa85caec6213105222e10e4900b225/runpod/serverless/worker.py#L29

therealadityashankar: https://github.com/runpod/runpod-python/compare/main...therealadityashankar:runpod-python:main#diff-4faab0673ed12ff33579213adc10043878e8079110e0762f29147d0c484270bbR82

this line specifically won't work without the future bit

`def run(self, endpoint_input : Any) -> Job`, because the Job class is defined after the type hint

justinmerrell: Oh, is this referring to the order it appears on the file? Can we move it up?

therealadityashankar: yes it is referencing the order it appears on the file, we can, and I've done it and removed the __future__ reference now

let me know if there's anything else I can improve and I'll be happy to do so

thanks

justinmerrell (approved):

PR #63, opened 2023-07-19
Enabling multi-job, single-worker in Runpod-Python.

Jorghi12:
This PR enabled multi-job single-worker support inside Runpod-Python and was recently tested successfully with the vLLM serverless worker. Let's start refactoring this implementation, if possible.

### DIFF
--- a/runpod/serverless/modules/rp_job.py
+++ b/runpod/serverless/modules/rp_job.py
@@ -71,7 +71,7 @@ async def get_job(session: ClientSession, retry=True) -> Optional[Dict[str, Any]
     return next_job
 
 
-def run_job(handler: Callable, job: Dict[str, Any]) -> Dict[str, Any]:
+async def run_job(handler: Callable, job: Dict[str, Any]) -> Dict[str, Any]:
     """
     Run the job using the handler.
     Returns the job output or error.
@@ -79,7 +79,7 @@ def run_job(handler: Callable, job: Dict[str, Any]) -> Dict[str, Any]:
     log.info(f'{job["id"]} | Started')
 
     try:
-        job_output = handler(job)
+        job_output = await handler(job)
         log.debug(f'{job["id"]} | Handler output: {job_output}')
 
         run_result = {"output": job_output}
--- a/runpod/serverless/modules/rp_job.py
+++ b/runpod/serverless/modules/rp_job.py
@@ -10,7 +10,7 @@ import traceback
 from aiohttp import ClientSession
 
 from runpod.serverless.modules.rp_logger import RunPodLogger
-from .worker_state import WORKER_ID
+from .worker_state import WORKER_ID, Jobs
 from .rp_tips import check_return_size
 
 JOB_GET_URL = str(os.environ.get('RUNPOD_WEBHOOK_GET_JOB')).replace('$ID', WORKER_ID)
@@ -29,9 +29,12 @@ async def get_job(session: ClientSession, retry=True) -> Optional[Dict[str, Any]
     Note: Retry True just for ease of, if testing improved this can be removed.
     """
     next_job = None
+    job_list = Jobs()
+
     while next_job is None:
         try:
-            async with session.get(JOB_GET_URL) as response:
+            job_in_progress = '1' if job_list.get_job_list() is not None else '0'
+            async with session.get(JOB_GET_URL + "&job_in_progress={}".format(job_in_progress)) as response:
                 if response.status == 204:
                     log.debug("No content, no job to process.")
                     if not retry:
@@ -117,7 +120,7 @@ def run_job(handler: Callable, job: Dict[str, Any]) -> Dict[str, Any]:
         return run_result  # pylint: disable=lost-exception
 
 
-def run_job_generator(
+async def run_job_generator(
         handler: Callable,
         job: Dict[str, Any]) -> Generator[Dict[str, Union[str, Any]], None, None]:
     '''
@@ -125,7 +128,7 @@ def run_job_generator(
     Yields output partials from the generator.
     '''
     try:
-        job_output = handler(job)
+        job_output = await handler(job)
         for output_partial in job_output:
             yield {"output": output_partial}
     except Exception as err:    # pylint: disable=broad-except
@@ -134,4 +137,4 @@ def run_job_generator(
     finally:
         log.info(f'{job["id"]} | Finished ')
 
-        return None  # pylint: disable=lost-exception
+        # return None  # pylint: disable=lost-exception
--- a/runpod/serverless/worker.py
+++ b/runpod/serverless/worker.py
@@ -169,6 +169,7 @@ async def run_worker_multi(config: Dict[str, Any]) -> None:
                 scalar.background_tasks.add(task)
                 task.add_done_callback(scalar.background_tasks.discard)
 
+        log.debug("rp_debugger | The event loop has closed due to {}, #2.")
         asyncio.get_event_loop().stop() # Stops the worker loop if the kill_worker flag is set.
 
 
@@ -241,5 +242,7 @@ def main(config: Dict[str, Any]) -> None:
                 asyncio.ensure_future(run_worker(config), loop=work_loop)
             work_loop.run_forever()
 
+        except Exception as e:  
+            log.debug("rp_debugger | The event loop has closed due to {}, #1.".format(e))
         finally:
             work_loop.close()
--- a/runpod/serverless/worker.py
+++ b/runpod/serverless/worker.py
@@ -94,6 +94,8 @@ class Scaler():
 
         # If our worker is fully utilized or the SLS queue is throttling, reduce the job query rate.
         if self.handler_fully_utilized() is True:
+            log.debug("The handler is fully utilized. Downscaling now.")
+
             # Reduce job query rate.
             self.num_concurrent_requests = int(max(
                 self.num_concurrent_requests // SCALE_FACTOR, MIN_CONCURRENT_REQUESTS))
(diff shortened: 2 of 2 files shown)


runpod/serverless/modules/rp_job.py
```diff
@@ -29,9 +31,11 @@ async def get_job(session: ClientSession, retry=True) -> Optional[Dict[str, Any]
     Note: Retry True just for ease of, if testing improved this can be removed.
     """
     next_job = None
+
     while next_job is None:
         try:
-            async with session.get(JOB_GET_URL) as response:
+            job_in_progress = '1' if job_list.get_job_list() else '0'
+            async with session.get(JOB_GET_URL + f"&job_in_progress={job_in_progress}") as response: # pylint: disable=line-too-long
```
justinmerrell: Let's move this into a local function `_job_get_url` and add a docstring with some context of what `job_in_progress` is doing.

runpod/serverless/worker.py
```diff
@@ -46,56 +46,234 @@ def _is_local(config) -> bool:
     return False
 
 
+class JobProcessor():
```
justinmerrell: The scaling control seems to be an auxiliary function of the worker and not part of the core loop logic. Would it be easier to follow if the scaling logic got moved into oa `rp_scale` module?

runpod/serverless/worker.py
```diff
@@ -46,56 +46,234 @@ def _is_local(config) -> bool:
...
+                    get_job(session, retry=False)
+                ) for _ in range(self.num_concurrent_get_job_requests)]
+
+            for job_future in asyncio.as_completed(tasks):
+                job = await job_future
+                self.job_history.append(1 if job else 0)
+
+                # Add the job to our list of jobs
+                if job:
+                    job_list.add_job(job["id"])
```
justinmerrell: Is there any downside to updating the job list within the same function where we get the job. So when we do `get_job` and a job is returned that means it has been added the list and is ready to process.

runpod/serverless/worker.py
```diff
@@ -46,56 +46,234 @@ def _is_local(config) -> bool:
...
-                job_result["output"]["rp_debugger"]["ready_delay_ms"] = ready_delay
-            else:
-                log.debug("rp_debugger | Flag not set, skipping debugger output.")
-                rp_debugger.clear_debugger_output()
+                # Send the job result to SLS
+                await send_result(session, job_result, job)
 
-            await send_result(session, job_result, job)
+                log.info(f'{job["id"]} | Finished')
+                job_list.remove_job(job["id"])
```
justinmerrell, line 109: Similar to appending the list when we `get_job`, can we handle removing the job from the list when the job is returned?

justinmerrell (changes requested):

Jorghi12:
Successfully handled first round of refactors.

runpod/serverless/worker.py
```diff
@@ -105,13 +126,15 @@ def main(config: Dict[str, Any]) -> None:
...
-        rp_local.run_local(config)
-
+        asyncio.run(rp_local.run_local(config))
     else:
         try:
             work_loop = asyncio.new_event_loop()
             asyncio.ensure_future(run_worker(config), loop=work_loop)
             work_loop.run_forever()
 
+        except Exception as exception:  # pylint: disable=broad-exception-caught
```
justinmerrell: My approach to catching exceptions has been to tighten up code that could result in an exception. Ideally, the only board catch exception we want in the worker would be related to the handler since that code comes from users and is outside our control. 

The danger with trying to catch exceptions proactively is misleading errors that make debugging difficult. However, I am open to hearing a different approach.

setup.cfg
```diff
@@ -24,7 +24,7 @@ classifiers =
 [options]
 include_package_data = true
 packages = find:
-python_requires = >= 3.10
+python_requires = >= 3.9
```
justinmerrell: Unless you changed things to be backward compatible, other sections of code will not work with 3.9 due to syntax that was introduced in 3.10

runpod/serverless/modules/rp_http.py
```diff
@@ -46,6 +48,9 @@ async def send_result(session, job_data, job):
         await transmit(session, job_data, job_done_url)
         log.debug(f"{job['id']} | Results sent.")
 
+        log.info(f'{job["id"]} | Finished')
```
justinmerrell: Nitpicking, but I try to place the log after the final action. In this case, "finished" should indicate the job is done and has been removed from the list. 

If, for whatever reason, the job is not removed then we have a misleading log.

runpod/serverless/modules/rp_job.py
```diff
@@ -29,9 +45,10 @@ async def get_job(session: ClientSession, retry=True) -> Optional[Dict[str, Any]
     Note: Retry True just for ease of, if testing improved this can be removed.
     """
     next_job = None
+
     while next_job is None:
         try:
-            async with session.get(JOB_GET_URL) as response:
+            async with session.get(_job_get_url()) as response: # pylint: disable=line-too-long
```
justinmerrell: `# pylint: disable=line-too-long` should only be used if absolutely necessary, otherwise refactor. 

However, since you refactored this does not appear to be necessary now.

runpod/serverless/modules/rp_job.py
```diff
@@ -134,4 +157,4 @@ def run_job_generator(
     finally:
         log.info(f'{job["id"]} | Finished ')
 
-        return None  # pylint: disable=lost-exception
+        # return None  # pylint: disable=lost-exception
```
justinmerrell: I can't remember why I had this structured like this, should be able to remove this line entirely if not required.

runpod/serverless/modules/rp_scale.py
justinmerrell, line 1: If the job processing should be included, let us re-think the file name; otherwise, we should re-factor so that only scale logic is part of this file.

justinmerrell (changes requested):

runpod/serverless/modules/rp_scale.py
Jorghi12: The entire file handles only scaling logic. I'll rename it to JobScaler.

runpod/serverless/modules/rp_job.py
Jorghi12: Agreed, that's a relic prior to refactoring.

runpod/serverless/modules/rp_http.py
Jorghi12: I like this philosophy. I'll update `send_result` log "finished" only after the job has been removed from the list (and of course, sent to the server successfully).

runpod/serverless/worker.py
Jorghi12: I'm neutral. I originally included this line to pipe exception information into the serverless logs on the main site; however, it's not needed anymore on my end. The philosophy makes sense. I'll go ahead and remove this line.

setup.cfg
Jorghi12: I've updated to 3.10 by making our serverless worker compatible with 3.10 now.

runpod/serverless/modules/rp_job.py
```diff
@@ -10,12 +11,27 @@
...
+    """
+    Prepare the URL for making a 'get' request to the serverless API (sls).
+
+    This function constructs the appropriate URL for sending a 'get' request to the serverless API,
+    ensuring that the request will be correctly routed and processed by the API.
+
+    Returns:
+        str: The prepared URL for the 'get' request to the serverless API.
+    """
+    job_in_progress = '1' if job_list.get_job_list() else '0'
```
justinmerrell, line 33: Did I see something in Discord related to this? Some scale system and not just bool?

runpod/serverless/worker.py
```diff
@@ -49,53 +49,77 @@ def _is_local(config) -> bool:
...
     """
-    Starts the worker loop.
+    Starts the worker loop for multi-processing.
+
+    Args:
+        config (Dict[str, Any]): Configuration parameters for the worker.
     """
     auth_header = _get_auth_header()
-    async with aiohttp.ClientSession(headers=auth_header, timeout=_TIMEOUT) as session:
+    connector = aiohttp.TCPConnector(limit=None)
```
justinmerrell, line 58: I wonder if we can control the number of jobs being requested by simply adjusting this parameter instead. 

Not something required now, but can keep it in mind.

runpod/serverless/worker.py
```diff
@@ -49,53 +49,77 @@ def _is_local(config) -> bool:
...
+            connector=connector, headers=auth_header, timeout=_TIMEOUT) as session:
 
         heartbeat.start_ping()
 
-        kill_worker = False # Flag to kill the worker after job is complete.
-        while kill_worker is False:
-            job = await get_job(session)
+        # Flag to kill the worker after job is complete.
+        job_scaler = JobScaler(
+            handler_fully_utilized=config.get('handler_fully_utilized'),
```
justinmerrell, line 67: We will test with `handler_fully_utilized` for now but am not sold on the name yet for the final release.

runpod/serverless/modules/rp_job.py
Jorghi12: It's for v1JobTake. Essentially, when job_in_progress == "1," it implies that the worker has jobs in progress. This is important for the multi-processing case to prevent premature downscaling.

runpod/serverless/worker.py
Jorghi12: Great question. One thing to be careful about is that the rate needs to adjust over time due to changes in workload distribution. But yes, creating a simpler parameterization of the concurrency would be very beneficial in improving maintainability.

runpod/serverless/worker.py
Jorghi12: Agreed, don't like the name either.

justinmerrell (approved):

PR #66, opened 2023-07-26
minor worker refactoring

justinmerrell:
I ran into the following error when running the main branch as is:

image

I refactored a bit to help debug.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 aiohttp >= 3.8.4
+
 backoff == 2.2.1
 boto3 >= 1.26.165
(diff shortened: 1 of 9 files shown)


Jorghi12:
You have to use the following because in asyncio, the wait() statement on an empty list isn't allowed.

```
if self.background_get_job_tasks:
      await asyncio.wait(self.background_get_job_tasks)
```

Jorghi12:
All tests pass now with full code coverage. I'll start testing vLLM now.

Jorghi12 (approved):
Approved :)

PR #70, opened 2023-07-28
Update rp_ping.py

justinmerrell:
Fixed the error catching but did not update the `status_code` [🤦](https://emojipedia.org/person-facepalming/)


Jorghi12 (approved):
LGTM, lint case

PR #76, opened 2023-07-31
Polish

justinmerrell:
While testing I noticed that ping would timeout and then not restart, fixed that and then fixed the corresponding tests.

### DIFF
--- a/runpod/serverless/modules/rp_ping.py
+++ b/runpod/serverless/modules/rp_ping.py
@@ -28,17 +28,23 @@ class HeartbeatSender:
             HeartbeatSender._instance = object.__new__(cls)
         return HeartbeatSender._instance
 
-    async def start_ping(self):
+    async def start_ping(self, test=False):
         '''
         Sends heartbeat pings to the Runpod server.
         '''
-        try:
-            while True:
-                await self._send_ping()
-                await asyncio.sleep(int(PING_INTERVAL / 1000))
-        except (aiohttp.ClientError, asyncio.TimeoutError, asyncio.CancelledError) as err:
-            log.error(f"Ping Error: {err}, attempting to restart ping.")
-            await self.start_ping()
+        while True:
+            try:
+                while True:
+                    await self._send_ping()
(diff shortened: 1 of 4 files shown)


runpod/serverless/modules/rp_ping.py
```diff
@@ -28,13 +28,23 @@ def __new__(cls):
             HeartbeatSender._instance = object.__new__(cls)
         return HeartbeatSender._instance
 
-    async def start_ping(self):
+    async def start_ping(self, test=False):
```
Jorghi12, line 31: That's pretty clever. I hadn't considered refactoring in this sort of way to improve tests.

tests/test_serverless/test_modules/test_ping.py
```diff
@@ -57,7 +57,7 @@ async def test_start_ping(self, mock_get_return):
         mock_session.headers.update = Mock()
 
         # Success case
-        await new_ping._send_ping() # pylint: disable=protected-access
+        await new_ping.start_ping(test=True)
```
Jorghi12, line 60: Great improvement. Tests should definitely use the exposed methods, and not methods meant to be private/protected.

tests/test_serverless/test_modules/test_ping.py
```diff
@@ -67,5 +67,5 @@ async def test_start_ping(self, mock_get_return):
         mock_get_return.side_effect = aiohttp.ClientError("Test Error")
 
         with patch("runpod.serverless.modules.rp_ping.log.error") as mock_log_error:
-            await new_ping._send_ping() # pylint: disable=protected-access
-            assert mock_log_error.call_count == 2
+            await new_ping.start_ping(test=True)
+            assert mock_log_error.call_count == 1
```
Jorghi12, line 71: Great improvement. Tests should definitely use the exposed methods, and not methods meant to be private/protected.

Jorghi12 (approved):
LGTM! If code coverage and tests fully pass, we can merge this.

PR #78, opened 2023-08-01
Support both async and non-async function generation.

Jorghi12:
This PR updates the detection of HTTP streaming inside the serverless worker by checking if the handler is either an async or non-async function generation handler. In the past, we only detected the non-async function generations, which were insufficient for full support.

### DIFF
--- a/runpod/serverless/worker.py
+++ b/runpod/serverless/worker.py
@@ -76,15 +76,13 @@ async def run_worker(config: Dict[str, Any]) -> None:
 
                 # If refresh_worker is set, pod will be reset after job is complete.
                 if config.get("refresh_worker", False):
-                    log.info(
-                        f"refresh_worker | Flag set, stopping pod after job {job['id']}.")
+                    log.info(f"refresh_worker | Flag set, stopping pod after job {job['id']}.")
                     job_result["stopPod"] = True
                     job_scaler.kill_worker()
 
                 # If rp_debugger is set, debugger output will be returned.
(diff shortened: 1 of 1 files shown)


runpod/serverless/worker.py
```diff
@@ -74,20 +76,25 @@ async def process_job(job):
 
                 # If refresh_worker is set, pod will be reset after job is complete.
                 if config.get("refresh_worker", False):
-                    log.info(f"refresh_worker | Flag set, stopping pod after job {job['id']}.")
+                    log.info(
```
justinmerrell: Is your text editor wrapping this line? I belive it should be less than 100 as is.

runpod/serverless/worker.py
```diff
@@ -74,20 +76,25 @@ async def process_job(job):
...
+                        f"refresh_worker | Flag set, stopping pod after job {job['id']}.")
                     job_result["stopPod"] = True
                     job_scaler.kill_worker()
 
                 # If rp_debugger is set, debugger output will be returned.
                 if config["rp_args"].get("rp_debugger", False) and isinstance(job_result, dict):
-                    job_result["output"]["rp_debugger"] = rp_debugger.get_debugger_output()
-                    log.debug("rp_debugger | Flag set, returning debugger output.")
+                    job_result["output"]["rp_debugger"] = rp_debugger.get_debugger_output(
```
justinmerrell: Same here, wrapping seems unnecessary

justinmerrell (changes requested):

justinmerrell (approved):

PR #82, opened 2023-08-02
Preserve the final job result's post condition in streaming scenario.

Jorghi12:
For serverless workers handling streaming, it makes sense to concatenate all the intermediate outputs into the final output to get our job done. I think this functionality can be handled either inside the Runpod-Python or SLS side, but for ease of implementation, incorporating it here makes sense.


justinmerrell:
Unless there is a use case, let us leave this out for now. We do not know what will be streamed or if it will end and do not want to just keep accumulating data indefinitely.

PR #84, opened 2023-08-04
Data Race in TakeJob

Jorghi12:
In our job acquisition logic, we need to ensure sequential execution rather than parallel execution, in order to prevent a potential data race within the TakeJob logic in SLS.

### DIFF
--- a/runpod/serverless/modules/rp_scale.py
+++ b/runpod/serverless/modules/rp_scale.py
@@ -99,17 +99,26 @@ class JobScaler():
             if not self.is_alive():
                 break
 
-            tasks = [
-                asyncio.create_task(
-                    get_job(session, retry=False)
-                ) for _ in range(self.num_concurrent_get_job_requests)]
+            sequential = True
+            tasks = []
+            if sequential:
+                for _ in range(self.num_concurrent_get_job_requests):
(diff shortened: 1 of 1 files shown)


runpod/serverless/modules/rp_scale.py
```diff
@@ -99,17 +99,26 @@ async def get_jobs(self, session):
...
-                    get_job(session, retry=False)
-                ) for _ in range(self.num_concurrent_get_job_requests)]
-
-            for job_future in asyncio.as_completed(tasks):
-                job = await job_future
-                self.job_history.append(1 if job else 0)
-
-                if job:
-                    yield job
+            sequential = True
```
justinmerrell: If you are setting this to True, when can it ever be false? This would make the if condition useless right?

justinmerrell:
> In our job acquisition logic, we need to ensure sequential execution rather than parallel execution, in order to prevent a potential data race within the TakeJob logic in SLS.

Does it make sense to handle on the worker side? If I am understanding correctly in a case where a worker can handle multiple jobs we will have >1 request out and SLS will try to respond to both with the same job potentially?

justinmerrell (approved):
It does make the loop easier to follow and reduces lines of code.

PR #86, opened 2023-08-08
Heartbeat thread

justinmerrell:
Switching ping back to threads, when attempting to use asyncio any handler that was blocking would result in ping also stopping.


Jorghi12 (approved):
LGTM

PR #88, opened 2023-08-09
Preserve the final job result's post condition in streaming scenario.

Jorghi12:
After conversing with Justin, we have opted to combine streaming and non-streaming functionalities into a unified worker image. Furthermore, the resulting output can be consolidated into a single job output, tentatively labeled as `submit_stream_on_end`.

### DIFF
--- a/runpod/serverless/worker.py
+++ b/runpod/serverless/worker.py
@@ -54,7 +54,8 @@ async def _process_job(job, session, job_scaler, config):
         job_result = {'output': []}
         async for stream_output in generator_output:
             if config.get('return_aggregate_stream', False):
-                job_result['output'].append(stream_output['output'])
+                if 'output' in stream_output:
+                    job_result['output'].append(stream_output['output'])
             await stream_result(session, stream_output, job)
     else:
         job_result = await run_job(config["handler"], job)
--- a/runpod/serverless/worker.py
+++ b/runpod/serverless/worker.py
@@ -67,16 +67,14 @@ async def run_worker(config: Dict[str, Any]) -> None:
             async def process_job(job):
                 if inspect.isgeneratorfunction(config["handler"]) \
                     or inspect.isasyncgenfunction(config["handler"]):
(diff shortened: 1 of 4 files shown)


runpod/serverless/worker.py
```diff
@@ -70,9 +70,14 @@ async def process_job(job):
                     job_result = run_job_generator(config["handler"], job)
 
                     log.debug("Handler is a generator, streaming results.")
+                    job_outputs = []
                     async for job_stream in job_result:
+                        if config["submit_stream_on_end"]:
```
justinmerrell: a slight name change, still open to suggestions but something that has less ambiguity with that the result will be "return_aggregate_stream"

runpod/serverless/worker.py
```diff
@@ -70,9 +70,14 @@ async def process_job(job):
                     job_result = run_job_generator(config["handler"], job)
 
                     log.debug("Handler is a generator, streaming results.")
+                    job_outputs = []
```
justinmerrell: job_results = {'output': []}

runpod/serverless/worker.py
```diff
@@ -70,9 +70,14 @@ async def process_job(job):
...
                     async for job_stream in job_result:
+                        if config["submit_stream_on_end"]:
+                            job_outputs.append(job_stream)
```
justinmerrell: job_results['output'].append(job_stream)

runpod/serverless/worker.py
```diff
@@ -70,9 +70,14 @@ async def process_job(job):
...
                         await stream_result(session, job_stream, job)
                     job_result = {}
```
justinmerrell: Can these lines now be eliminated?

runpod/serverless/worker.py
Jorghi12: return_aggregate_stream is a better name. sg.

justinmerrell (approved):

PR #107, opened 2023-08-25
Enable safe, parallel jobTake inside single-processing / multi-processing logic.

Jorghi12:
The JobTake endpoint within SLS is not safe with respect to the "job in progress." We are updating Runpod Python to oscillate between sequential and parallel processing when it makes the most sense. This update will allow for a high jobTake uptake for multi-processing workloads (e.g. vLLM) without leading to unsafe situations inside SLS.


justinmerrell:
Rebase required.

PR #114, opened 2023-08-28
Fix streaming for LLMs

Jorghi12:
The LLM endpoints require that we use streaming and additionally the final streamed output should be in the job done output. This enables this functionality.


justinmerrell:
Let's define the docs on the expected behavior of how streaming should be handled and where functionality is divided between the SDK and the handler. 

If the stream looks like `A` > `B` > `C`, then we will only be returning `C` as well as omitting the empty return at the end. 

If the stream looks like `A` > `AB` > `ABC`, then this makes sense, however, that would be making an assumption that all users who implement streaming will have that behavior from their handler.

PR #130, opened 2023-09-14
explicit region_name in boto3 client init

MartinKlefas:
To make sure I'm targeting the right region (specifically, eu-west-2), I've explicitly set the `region_name` parameter during the boto3 client initialization. Without this setting, there's a risk the client could default to `us-east-1` or pick up a region from system AWS configuration files.

By being explicit about the `region_name`, I've ensured the client recognizes the intended region, which should lead to the generation of presigned URLs that correctly match the desired AWS zone. I've tried to read the region from the bucket URL and tried to do that for a variety of common providers. It could also be addressed via an environment variable if that brings it inline better with your design philosophy

This PR addresses #126

### DIFF
--- a/runpod/serverless/utils/rp_upload.py
+++ b/runpod/serverless/utils/rp_upload.py
@@ -23,6 +23,17 @@ logger = logging.getLogger("runpod upload utility")
 FMT = "%(filename)-20s:%(lineno)-4d %(asctime)s %(message)s"
 logging.basicConfig(level=logging.INFO, format=FMT, handlers=[logging.StreamHandler()])
 
+def extract_region_from_url(endpoint_url):
+    # AWS/backblaze S3-like URL
+    if '.s3.' in endpoint_url:
+        return endpoint_url.split('.s3.')[1].split('.')[0]
+    # DigitalOcean Spaces-like URL
+    elif '.digitaloceanspaces.com' in endpoint_url:
+        return endpoint_url.split('.')[1].split('.digitaloceanspaces.com')[0]
+    else:
+        # Additional cases can be added here
+        return None
+  
 
 # --------------------------- S3 Bucket Connection --------------------------- #
 def get_boto_client(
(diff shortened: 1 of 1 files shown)


justinmerrell:
May you address the `Code scanning results / CodeQL ` results?

MartinKlefas:
> May you address the `Code scanning results / CodeQL ` results?

I've done that now - the code still fails the pytest though, as "https://your-bucket-endpoint-url.com'" doesn't contain a valid s3 compliant URL and so there's no region to be found and submitted in the boto3 client initialisation. 

I don't know if it's possible to change pytest so that it submits a compliant dummy url: https://testbucket.s3.testregion.amazonaws.com  or something?

justinmerrell:
Thank you for the correction, I see that I need to add the trigger to run that workflow on PRs as well. 

Do you have screenshots of the exact test error?

MartinKlefas:
Please see below:
image
I suppose a good workaround would be to return the default region from env if the use case isn't found, rather than None?

justinmerrell:
Oh, I see, it is just the assertion test that needs to be updated to include "region_name=None" for the expected call.

MartinKlefas:
> Oh, I see, it is just the assertion test that needs to be updated to include "region_name=None" for the expected call.

Is that something I should be doing? Sorry I'm a bit new to all this!

justinmerrell:
Feel free to take a crack at it; otherwise, I'd be glad to take the PR over the finish line. Your contribution is already greatly appreciated! 

In the rp_upload test file there should be something like "assert called with" and then has the client input but will be missing the `region_name`

MartinKlefas:
> Feel free to take a crack at it; otherwise, I'd be glad to take the PR over the finish line. Your contribution is already greatly appreciated!
> 
> In the rp_upload test file there should be something like "assert called with" and then has the client input but will be missing the `region_name`

I've made those changes now, but not sure how to get that other end-to-end test to run - the error coming up is about docker usernames:
Run docker/login-action@v3
Error: Username and password required
Not sure how or where to put in dummy ones to get the test to pass I'm afraid.

justinmerrell:
That last test needs to be run from our account. PR looks good; thanks for the submission!

PR #160, opened 2023-10-05
Added code to handle non-image uploads in runpod.serverless.utils.rp_upload.files(...)

MartinKlefas:
The existing upload code assumes that the user is trying to upload an image, and forces the S3 content type to PNG regardless of what the file actually is.

This version first tries an image-based approach, setting the content type by the image type, and then falls back to a file-based upload. 

Any exceptions caused by anything other than PIL detecting it's not an image will still be handled in the remainder of the code.

### DIFF
--- a/runpod/serverless/utils/rp_upload.py
+++ b/runpod/serverless/utils/rp_upload.py
@@ -21,7 +21,6 @@ from botocore.config import Config
 from tqdm_loggable.auto import tqdm
 
 
-
 logger = logging.getLogger("runpod upload utility")
 FMT = "%(filename)-20s:%(lineno)-4d %(asctime)s %(message)s"
 logging.basicConfig(level=logging.INFO, format=FMT, handlers=[logging.StreamHandler()])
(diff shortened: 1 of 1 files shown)


runpod/serverless/utils/rp_upload.py
```diff
@@ -14,13 +14,14 @@
...
 import boto3
-from PIL import Image
+from PIL import Image, UnidentifiedImageError
 from boto3 import session
 from boto3.s3.transfer import TransferConfig
 from botocore.config import Config
 from tqdm_loggable.auto import tqdm
 
 
+
```
justinmerrell: Everything else looks good, I almost feel bad asking for the one change to get rid of this extra blank line :)

justinmerrell (changes requested):

justinmerrell:
Solid contribution, just a super minor nit-picking change request and then it should be good to merge :)

MartinKlefas:
I'd say a solid 50% of my code base is extra blank lines - happy to remove one or two!

PR #266, opened 2024-01-12
Adding a send and download to runpod pod cli

justinwlin:
A common issue I see is with people not having SCP working or trying to get it to work. Yes, I could give them a whole tutorial on SCP. Or I could tell them, at least do this one time, and in the future, you'll never have to do this again.

So I added a SEND / DOWNLOAD command, and tried to add helpful --helps so that people have examples and so that after things are sent it gives them absolute paths and CD commands.

Gave it a try myself, and seems to work well... I just utilized essentially what you guys had from the SSHConnection class 🙏🏼 so feel free to give a try and make corrections if needed.

P.S if this does get merged, can you use the `squash and merge` button on github this so the commit history is clean ;D. Thanks~

### DIFF
--- a/runpod/cli/groups/pod/commands.py
+++ b/runpod/cli/groups/pod/commands.py
@@ -93,7 +93,7 @@ def send_file(pod_id, local_path, remote_path):
         click.echo(
             f"File sent successfully to {remote_directory_display} on pod {pod_id}."
         )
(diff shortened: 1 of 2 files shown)


justinmerrell:
This is a good idea; I am OK with adding this, I need it to pass the CI/CD checks. You are welcome to take a stab at it, I just won't have time to get around to doing so for a bit.

PR #289, opened 2024-02-01
Runs black, isort, black on docs

rachfop:


runpod/__init__.py
```diff
@@ -1,32 +1,19 @@
...
-from .version import __version__
-from .api.ctl_commands import (
-    get_user, update_user_settings,
-    get_gpu, get_gpus,
-    get_pod, get_pods, create_pod, stop_pod, resume_pod, terminate_pod,
-    create_template,
-    get_endpoints, create_endpoint, update_endpoint_template
-)
-from .cli.groups.config.functions import set_credentials, check_credentials, get_credentials
-
```
rachfop, line 19: Seems like these were mostly unused.

justinmerrell: They are exposed to allow:

```python
import runpod

runpod.get_pods()
runpod.create_endpoint()
```

It is possible the implementation was incorrect, if there is a more pythonic way to implement please let me know.

justinmerrell:
The changes appear to prevent current actions from passing, current tests could be wrong but will need all to pass before merging.

PR #294, opened 2024-02-09
feat: added runpod command to create container registry auth and rena…

### LINKED ISSUE #283: Implement methods for credentials and template creation
Hi! I've faced the next issue trying to automate template and pod creation: it's impossible to create container registry credentials from SDK which I want to use in a future to create a template.

So the desired features I would ask you to implement:
- add method to create credentials:

`runpod.create_container_registry_credentials(name, username, password): ->
        return container_registry_auth_id
`
- add additional attribute to provide `container_registry_auth_id` into `runpod.create_template()` method

I hope it would be useful not only for me but other developers. Anyway, thanks for your work!

yamanahlawat:
…med registry_auth to registry_auth_id to avoid confusion

fixes #283 

@justinmerrell

### DIFF
--- a/examples/api/create_container_registry_auth.py
+++ b/examples/api/create_container_registry_auth.py
@@ -1,0 +1,19 @@
+""" Example of creating a container registry auth with the Runpod API. """
+
+import runpod
+
+# Set your global API key with `runpod config` or uncomment the line below:
+# runpod.api_key = "YOUR_RUNPOD_API_KEY"
+
+try:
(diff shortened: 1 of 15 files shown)


yamanahlawat:
update test cases and lint errors

yamanahlawat:
@justinmerrell  Thanks for looking into it.

Additionally, we need to display the `registry_auth_id` in the user interface, enabling users to easily copy it for use in creating templates automatically.

PR #315, opened 2024-05-21
Update Readme

rachfop:
Some users reporting problems with GeForce GPUs from the Python SDK. 
Both Pods start for me, but will update the reademe. 

Screenshot 2024-05-21 at 12 48 48 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ pods = runpod.get_pods()
 pod = runpod.get_pod(pod.id)
 
 # Create a pod
-pod = runpod.create_pod("test", "runpod/stack", "NVIDIA GeForce RTX 3070")
+pod = runpod.create_pod("test", "runpod/stack", "NVIDIA RTX A6000")
 
 # Stop the pod
 runpod.stop_pod(pod.id)


README.md
```diff
@@ -138,7 +138,7 @@ pods = runpod.get_pods()
 pod = runpod.get_pod(pod.id)
 
 # Create a pod
-pod = runpod.create_pod("test", "runpod/stack", "NVIDIA GeForce RTX 3070")
+pod = runpod.create_pod("test", "runpod/stack", "NVIDIA RTX A6000")
```
ashleykleynhans, line 141: What exactly is the point in changing this? It offers no benefit and does not fix any typos or misinformation.  PRs should only be created if you have something meaningful to contribute.

ashleykleynhans (changes requested):

README.md
rachfop: Users reported that `NVIDIA GeForce RTX 3070` doesn't run. 
I tested `NVIDIA RTX A6000` and it works, but so did `NVIDIA GeForce RTX 3070`.

PR #329, opened 2024-07-25
Observability: implement tracing for http requests

deanq:
The following changes are made:
- `runpod.tracer` for instrumenting aiohttp.ClientSession and requests.Session. This will be used for future changes (OpenTelemetry instrumentation)
- `runpod.http_client` abstracts aiohttp.ClientSession and requests.Session direct calls for better control and modularity to prepare for future changes (using httpx)
- `runpod.http_client.AsyncClientSession` for async call is (temporarily) a factory method that returns a pre-configured aiohttp.ClientSession that is instrumented for tracing
- Request header: `X-Request-ID` holds Job ID for tracing purposes
- Support for env var overrides: `RUNPOD_API_BASE_URL` & `RUNPOD_ENDPOINT_BASE_URL`
- New logger level: `TRACE` value for env var `RUNPOD_LOG_LEVEL`
- Misc. corrections for pylint, unittest and build

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ backoff >= 2.2.1
 boto3 >= 1.26.165
 click >= 8.1.7
 colorama >= 0.2.5, < 0.4.7
+cryptography < 43.0.0
 fastapi[all] >= 0.94.0
 paramiko >= 3.3.1
 prettytable >= 3.9.0
--- a/runpod/http_client.py
+++ b/runpod/http_client.py
@@ -1,0 +1,89 @@
+"""
+HTTP Client abstractions
+"""
+
+import os
+import aiohttp
+import requests
+from .tracer import (
+    get_aiohttp_tracer,
+    get_request_tracer,
+)
+from .cli.groups.config.functions import get_credentials
+from .user_agent import USER_AGENT
+
+
+def get_auth_header():
+    """
+    Produce a header dict with the `Authorization` key derived from
+    credentials.get("api_key") OR os.getenv('RUNPOD_AI_API_KEY')
+    """
+    if credentials := get_credentials():
+        auth = credentials.get("api_key")
+    else:
+        auth = os.getenv("RUNPOD_AI_API_KEY")
+
+    return {
+        "Content-Type": "application/json",
+        "Authorization": f"{auth}",
+        "User-Agent": USER_AGENT,
+    }
+
+
+class AsyncClientSession(aiohttp.ClientSession):
+    """
+    Inherits aiohttp.ClientSession and overrides with our preferred params
+    TODO: use httpx
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            connector=aiohttp.TCPConnector(limit=0),
+            headers=get_auth_header(),
+            timeout=aiohttp.ClientTimeout(600, ceil_threshold=400),
+            trace_configs=[get_aiohttp_tracer()],
+            *args,
+            **kwargs,
+        )
+
+
+class SyncClientSession(requests.Session):
+    """
+    Inherits requests.Session to override `request()` method for tracing
+    TODO: use httpx
+    """
+
+    def request(self, method, url, **kwargs):  # pylint: disable=arguments-differ
+        """
+        Override for tracing. Not using super().request()
+        to capture metrics for connection and transfer times
+        """
+        with get_request_tracer() as tracer:
+            # Separate out the kwargs that are not applicable to `requests.Request`
+            request_kwargs = {
+                k: v
+                for k, v in kwargs.items()
+                if k in requests.Request.__init__.__code__.co_varnames
+            }
+            send_kwargs = {k: v for k, v in kwargs.items() if k not in request_kwargs}
+
+            # Create a PreparedRequest object to hold the request details
+            req = requests.Request(method, url, **request_kwargs)
+            prepped = self.prepare_request(req)
+            tracer.request = prepped  # Assign the request to the tracer
+
+            # Merge environment settings
+            settings = self.merge_environment_settings(
+                prepped.url,
+                send_kwargs.get("proxies"),
+                send_kwargs.get("stream"),
+                send_kwargs.get("verify"),
+                send_kwargs.get("cert"),
+            )
+            send_kwargs.update(settings)
+
+            # Send the request
+            response = self.send(prepped, **send_kwargs)
+            tracer.response = response  # Assign the response to the tracer
+
+            return response
--- a/runpod/http_client.py
+++ b/runpod/http_client.py
@@ -3,8 +3,12 @@ HTTP Client abstractions
 """
 
 import os
-import aiohttp
 import requests
+from aiohttp import (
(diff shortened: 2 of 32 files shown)


runpod/http_client.py
```diff
@@ -0,0 +1,92 @@
...
+    credentials.get("api_key") OR os.getenv('RUNPOD_AI_API_KEY')
+    """
+    if credentials := get_credentials():
+        auth = credentials.get("api_key")
+    else:
+        auth = os.getenv("RUNPOD_AI_API_KEY")
+
+    return {
+        "Content-Type": "application/json",
+        "Authorization": f"{auth}",
```
ef0xa: Trivial: couldn't this jsut be `auth` without the f-string?

runpod/http_client.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Override for tracing. Not using super().request()
+        to capture metrics for connection and transfer times
+        """
+        with get_request_tracer() as tracer:
+            # Separate out the kwargs that are not applicable to `requests.Request`
+            request_kwargs = {
+                k: v
+                for k, v in kwargs.items()
+                if k in requests.Request.__init__.__code__.co_varnames
```
ef0xa, line 70: This is some deep python magic and could use some explanation. I don't understand it.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+    context.exception = str(params.exception)
+    elapsed = asyncio.get_event_loop().time() - context.on_request_start
+    context.transfer = elapsed - context.connect
+    context.end_time = time()
+
+    # log to error level
+    report_trace(context, params, elapsed, log.error)
+
+
+def report_trace(context: types.SimpleNamespace, params, elapsed, logger=log.trace):
```
ef0xa: this could use a documentation comment.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+async def on_response_chunk_received(
+    session, context, params: TraceResponseChunkReceivedParams
+):
+    if not hasattr(context, "response_size_bytes"):
+        context.response_size_bytes = 0
+    context.response_size_bytes += len(params.chunk)
+
+
+async def on_request_end(session, context, params: TraceRequestEndParams):
+    elapsed = asyncio.get_event_loop().time() - context.on_request_start
```
ef0xa, line 132: this could use a documentation comment.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+async def on_request_chunk_sent(session, context, params: TraceRequestChunkSentParams):
+    if not hasattr(context, "payload_size_bytes"):
+        context.payload_size_bytes = 0
+    context.payload_size_bytes += len(params.chunk)
+
+
+async def on_response_chunk_received(
+    session, context, params: TraceResponseChunkReceivedParams
```
ef0xa: this could use a documentation comment.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
+# pylint: disable-all
```
ef0xa, line 1: this could use a documentation string at the top

requirements.txt
```diff
@@ -5,6 +5,7 @@ backoff >= 2.2.1
 boto3 >= 1.26.165
 click >= 8.1.7
 colorama >= 0.2.5, < 0.4.7
+cryptography < 43.0.0
```
ef0xa, line 8: This seems a bit shady. Can we pin this more neatly?

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+        context.request_id = headers.get("x-request-id")
+        context.user_agent = headers.get("user-agent")
+
+    return context
+
+
+# Tracer for aiohttp
+
+
+async def on_request_start(session, context, params: TraceRequestStartParams):
```
ef0xa: please annotate the types here.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+        context.retries = context.trace_request_ctx["current_attempt"]
+
+
+async def on_connection_create_end(
+    session, context, params: TraceConnectionCreateEndParams
+):
+    context.connect = asyncio.get_event_loop().time() - context.on_request_start
+
+
+async def on_connection_reuseconn(
```
ef0xa, line 95: please annotate the types here.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+    context.on_request_start = asyncio.get_event_loop().time()
+    context.method = params.method
+    context.url = params.url.human_repr()
+    context.mode = "async"
+
+    if hasattr(context, "trace_request_ctx") and context.trace_request_ctx:
+        context.retries = context.trace_request_ctx["current_attempt"]
+
+
+async def on_connection_create_end(
```
ef0xa, line 86: add a doc string explaining what this actually does (or give it a better name)

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+async def on_request_end(session, context, params: TraceRequestEndParams):
+    elapsed = asyncio.get_event_loop().time() - context.on_request_start
+    context.transfer = elapsed - context.connect
+    context.end_time = time()
+
+    # log to trace level
+    report_trace(context, params, elapsed)
+
+
+async def on_request_exception(session, context, params: TraceRequestExceptionParams):
```
ef0xa: add a doc string explaining what it actually does.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+    if hasattr(context, "trace_request_ctx"):
+        delattr(context, "trace_request_ctx")
+
+    if hasattr(params, "response") and params.response:
+        context.response_status = params.response.status
+
+    logger(json.dumps(vars(context)), context.request_id)
+
+
+def get_aiohttp_tracer() -> TraceConfig:
```
ef0xa: Pedantic (naming): This makes a _new_ tracer, so "get" seems a bit wrong.

runpod/tracer.py
```diff
@@ -0,0 +1,194 @@
...
+    trace_config.on_request_end.append(on_request_end)
+    trace_config.on_request_exception.append(on_request_exception)
+
+    return trace_config
+
+
+# Tracer for requests
+
+
+class TraceRequest:
```
ef0xa, line 219: This could use a doc string explaining what this is for (and that it's a context manager).

ef0xa (approved):
I don't see any merge-blocking problems here, but please take the various review comments into consideration before you click merge. Good work.

requirements.txt
deanq: Not in this case. We're not using cryptography library directly. It's a dependency on other dependencies. In different Python versions, this could be different versions. We're capping it at before v43.0.0 because that's where the known deprecation error happens. Pegging it strictly could introduce unforeseen side-effects that could result in more bugs. As far as we know, prior to v43.0.0, it was working fine.

See https://github.com/paramiko/paramiko/issues/2419

PR #343, opened 2024-09-05
fix: remove pylint from CI

ef0xa:
> We should choose to merge _exactly one_ of https://github.com/runpod/runpod-python/pull/343 and https://github.com/runpod/runpod-python/pull/346

This pylint check is far too pedantic and often just wrong. It is reasonable, for instance, for some functions (rare ones!) to have 7+ arguments. 

[It also works on some code that is semantically identical to other code that fails: that's too flaky.](https://github.com/runpod/runpod-python/pull/342)

Drop it from CI.

Additionally, all the CI steps ran twice - once for the push, once for the pull. That's silly.


deanq:
@ef0xa 
I would not remove it since this is a publicly-maintained repo. We should still have some pylint check. I agree it's too much and unreasonable at some of its rules. I probed our current files and ranked the top 20 most common `pylint disable` comments here. These should just be disabled globally since it's obviously ignored by now:
```
 120 line-too-long
  66 import-outside-toplevel
  43 too-many-arguments
  41 unused-argument
  41 broad-except
  37 invalid-name
  34 protected-access
  23 broad-exception-raised
  22 too-few-public-methods
  19 W0613
  17 broad-exception-caught
  10 too-many-locals
  10 too-many-instance-attributes
  10 too-many-branches
  10 no-member
   8 unused-import
   8 signature-differs
   8 no-self-use
   8 import-error
   8 duplicate-code
```

I can suggest a `.pylintrc` file that will do the job

ef0xa:
@deanq , I'm fine with either way, but what you added to my PR adds a big `.pylintrc` _and_ still leaves them disabled. Let's do one or the other.

Issue #368, opened 2024-10-16
Version 1.7.3 is completely broken: High amount of failed and retried requests


ashleykleynhans:
**Describe the bug**
After upgrading to version 1.7.3, the number of retried and failed requests spiked dramatically - Of 155 total requests, 17 were retried and 11 failed. 

This is not happening with version 1.7.2, and things are back to normal after reverting to version 1.7.2.

**To Reproduce**
1. Use a ComfyUI serverless worker.
2. Ensure that `runpod` python module is version 1.7.3.
3. Send a few dozen requests to the endpoint.
4. Observe a high number of retried and failed requests.

**Expected behavior**
There should be NO retried or failed requests.

**Screenshots**
<img width="1290" alt="Screenshot 2024-10-16 at 08 33 39" src="https://github.com/user-attachments/assets/5812e2e5-cb41-470b-8a3e-7d2905f46492">

**Desktop (please complete the following information):**
N/A - Specific to RunPod Serverless.

**Smartphone (please complete the following information):**
N/A - Specific to RunPod Serverless.

**Additional context**
None.

PR #370, opened 2024-10-24
fix: long-running jobs crash with SIGTERM

deanq:
The bug was caused by an asyncio event loop locking up. Adding `finally: await asyncio.sleep(0)` fixed it.

This fixes issues like https://github.com/runpod/runpod-python/issues/368 and other similar reported issues.


Yhlong00 (approved):
👍

PR #372, opened 2024-11-03
Fix: is_generator logic

gabewillen:
Fix the is_generator logic to allow passing a callable object as a handler. 

```python
# simple example
import runpod


class Handler:

    def __call__(self, job):
        for x in range(10):
            yield {"message": f"Hello, World! {x}"}


runpod.serverless.start(
    {
        "handler": Handler(),
    }
)
```


gabewillen:
Closing this as it's not absolutely necessary

PR #376, opened 2024-11-16
Fix: failed requests due to race conditions in the job queue vs job progress

deanq:
Making the JobsProgress set asyncio-safe fixes the race condition issue


Yhlong00 (dismissed):
LGTM

PR #383, opened 2024-12-05
Fix: JobScaler issues that cause request failures

deanq:
- Integrated asyncio.Queue within JobScaler (removes JobsQueue) and fully take advantage of its blocking .get .put functions
- Using asyncio.Queue(maxsize) to dictate concurrency (via concurrency_modifier)
- JobScaler.set_scale() adjusts concurrency when needed and safe in runtime
- JobScaler.current_occupancy() uses asyncio.Queue size and JobsProgress(set) size to gate capacity
- Simpler/cleaner job acquisition steps
- Removed legacy tracers for http clients
- Run `python -m tests.test_serverless.test_modules.run_scale` locally to test and benchmark JobScaler features


Yhlong00 (approved):

gabewillen (commented):
This looks good to me!

PR #398, opened 2025-01-31
fix: better exception handling.

Yhlong00:
Just catch the `asyncio.TimeoutError` and throw in a debug log for it.

image

### DIFF
--- a/runpod/serverless/modules/rp_scale.py
+++ b/runpod/serverless/modules/rp_scale.py
@@ -211,7 +211,7 @@ class JobScaler:
                 )
             finally:
                 # Yield control back to the event loop
-                await asyncio.sleep(0.1)
+                await asyncio.sleep(0)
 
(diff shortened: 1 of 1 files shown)


runpod/serverless/modules/rp_scale.py
```diff
@@ -211,7 +211,7 @@ async def get_jobs(self, session: ClientSession):
                 )
             finally:
                 # Yield control back to the event loop
-                await asyncio.sleep(0)
+                await asyncio.sleep(0.1)
```
justinmerrell: Why the additional delay? 
https://superfastpython.com/what-is-asyncio-sleep-zero/

Yhlong00: Thanks for catching this! I was thinking about avoiding excessive API calls. That article was really helpful.

PR #404, opened 2025-03-20
Update async job streaming stop condition and test cases

zealotjin:
The async job's streaming method runs indefinitely without specifically calling a break outside of the function. 
This change handles the break when the job is complete AND all the stream data has been consumed. (The stop condition should check for the `{status: "COMPLETED", stream: []}` job response.

This is in sync with the runner stream implementation. The test case has also been updated.

### DIFF
--- a/runpod/endpoint/asyncio/asyncio_runner.py
+++ b/runpod/endpoint/asyncio/asyncio_runner.py
@@ -1,4 +1,4 @@
-""" Module for running endpoints asynchronously. """
+"""Module for running endpoints asynchronously."""
 
 # pylint: disable=too-few-public-methods,R0801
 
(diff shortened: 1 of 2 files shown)


runpod/endpoint/asyncio/asyncio_runner.py
```diff
@@ -89,9 +89,14 @@ async def stream(self) -> Any:
         while True:
             await asyncio.sleep(1)
             stream_partial = await self._fetch_job(source="stream")
-            if stream_partial["status"] not in FINAL_STATES:
+            if (
+                stream_partial["status"] not in FINAL_STATES
+                or len(stream_partial["stream"]) > 0
```
Yhlong00: or len(stream_partial.get("stream", [])) > 0

Yhlong00 (commented):
looks good to me

Yhlong00 (approved):

deanq (approved):

Issue #405, opened 2025-04-05
Linux/Windows File causes Pyinstaller Breakage


nakul-malhotra:
**Describe the bug**
“The published runpod package includes a Linux-only .so file (sls_core.so) that causes PyInstaller to break on macOS. Consider publishing a platform-specific wheel or excluding that file from the universal distribution.”

Here’s a filled-out version of the GitHub issue template tailored to your RunPod + PyInstaller situation:

⸻

To Reproduce
	1.	Create a Python project on macOS with a virtual environment
	2.	Install runpod via pip

pip install runpod

	3.	Attempt to build the project with PyInstaller (e.g. with --onefile flag)
	4.	Build fails with ValueError: Unknown Mach-O header: 0x7f454c46 when processing sls_core.so

Expected behavior
Expected PyInstaller to package the project without error. The .so file sls_core.so appears to be a Linux-only binary (ELF format) and is not compatible with macOS builds. Ideally, the package should:
	•	Exclude platform-specific .so files from universal wheels
	•	Or provide per-platform wheels (e.g., via --plat-name)
	•	Or ensure sls_core.so is only installed/downloaded on compatible platforms

Screenshots

file venv/lib/python3.11/site-packages/runpod/serverless/sls_core.so
# Output:
# ELF 64-bit LSB shared object, x86-64 ...

ValueError: Unknown Mach-O header: 0x7f454c46 in <_io.BufferedReader name='/Users/$USER/.../sls_core.so'>

Desktop (please complete the following information):
	•	OS: macOS 14.4.1 (Apple Silicon - arm64)
	•	Python Version: 3.11.11 (Homebrew)
	•	PyInstaller Version: 6.6.0
	•	runpod Version: 1.7.7

Smartphone (please complete the following information):
N/A

Additional context
This breaks PyInstaller builds on macOS because it attempts to analyze an ELF binary that’s not valid for the host system. While we can manually exclude the .so or patch around it, this makes runpod non-portable out of the box.

Suggested solutions:
	•	Publish platform-specific wheels, or
	•	Dynamically generate .so files only when needed on supported platforms, or
	•	Provide a Python fallback path for environments where native extensions aren’t supported

Happy to help test a fix!

nakul-malhotra:
update, seems like we cannot build at all because there is a reference: 

[2025-04-05][22:50:29][app_lib::process][INFO] Waiting for Python to start (elapsed: 50s)...
[...]

PR #406, opened 2025-04-08
Retire sls-core since we're no longer developing it

deanq:


deanq:
This fixes the issue https://github.com/runpod/runpod-python/issues/405

Yhlong00 (approved):

Issue #407, opened 2025-04-10
rp_api_concurrency > 1 doesn't work | managed to get it working


DifferentialityDevelopment:
**Describe the bug**
Running a local worker with API concurrency > 1 doesn't work
I first tried getting this to work in windows to no avail, I get the error 
WARNING:  You must pass the application as an import string to enable 'reload' or 'workers'
After a couple of hours of struggling I managed to get it working, but had to modify some code in the runpod package.

runpod/serverless/modules/rp_fastapi.py

Before:
```
def start_uvicorn(self, api_host="localhost", api_port=8000, api_concurrency=1):
        """
        Starts the Uvicorn server.
        """
        uvicorn.run(
            self.rp_app,
            host=api_host,
            port=int(api_port),
            workers=int(api_concurrency),
            log_level=os.environ.get("UVICORN_LOG_LEVEL", "info"),
            access_log=False,
        )
```

After:
```
def start_uvicorn(self, api_host="localhost", api_port=8000, api_concurrency=1):
        """
        Starts the Uvicorn server.
        """
        if api_concurrency > 1:
            # For multiple workers, we need to use the module:app format
            import uvicorn.workers
            uvicorn.run(
                "runpod.serverless.modules.rp_fastapi:app",
                host=api_host,
                port=int(api_port),
                workers=int(api_concurrency),
                log_level=os.environ.get("UVICORN_LOG_LEVEL", "info"),
                access_log=False,
                factory=True
            )
        else:
            # For single worker, we can use the app instance directly
            uvicorn.run(
                self.rp_app,
                host=api_host,
                port=int(api_port),
                workers=1,
                log_level=os.environ.get("UVICORN_LOG_LEVEL", "info"),
                access_log=False
            )
```

This finally got me a little further, upon which I realized I needed gunicorn installed, after installing I realized it doesn't work on windows, tried it out on my Linux machine and it finally worked as expected.

**To Reproduce**
1. Setup an basic worker as per the runpod docs here https://docs.runpod.io/serverless/workers/development/concurrency
2. Run an development api server with python main.py --rp_serve_api --rp_api_concurrency 4

**Expected behavior**
It should work without having to edit the code of the runpod package

**Desktop (please complete the following information):**
 - OS: Windows & Linux (Ubuntu 24.04)
 - Browser: N/a
 - Version 1.7.9 (latest)

**Additional context**

PR #408, opened 2025-04-10
Fix support for API concurrency > 1 by updating Uvicorn launch logic

DifferentialityDevelopment:
Fix for the bug when launching the serverless worker locally with api concurrency > 1

I got the following error:
WARNING:  You must pass the application as an import string to enable 'reload' or 'workers'.

As per my issue I created here
https://github.com/runpod/runpod-python/issues/407

### DIFF
--- a/runpod/serverless/modules/rp_fastapi.py
+++ b/runpod/serverless/modules/rp_fastapi.py
@@ -271,14 +271,28 @@ class WorkerAPI:
         """
         Starts the Uvicorn server.
         """
-        uvicorn.run(
-            self.rp_app,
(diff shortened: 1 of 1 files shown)


DifferentialityDevelopment:
I notice that running with concurrency is totally not a thing on Windows..
I was able to get it "working" on linux with the above changes
But it requires I give it test input data else the workers just exit immediately, if I just use concurrency = 1 then the worker persists.

I was going through all this to be able to stress test my application and gather data on per request memory usage, peak memory usage with different concurrency levels etc

PR #416, opened 2025-05-01
fix: parse Content-Disposition properly, closes #414

### LINKED ISSUE #414: download_files_from_urls, is not using the file extension from the Content-Disposition
**Describe the bug**
The code that uses email parser of header is not correct, therefore, the script won't use the original file extension.
https://github.com/runpod/runpod-python/blob/main/runpod/serverless/utils/rp_download.py#L58
Image

**To Reproduce**
```python
from email.parser import Parser
message = Parser().parsestr('content-disposition: inline; filename="1746011246.12130031.mp3"')
params = dict(message.items())
    
print(params)
```

Image

**Expected behavior**
It should parse the header in a different way.

felixmosh:
This PR fixes #414

### DIFF
--- a/runpod/serverless/utils/rp_download.py
+++ b/runpod/serverless/utils/rp_download.py
@@ -11,8 +11,7 @@ import re
 import uuid
 import zipfile
 from concurrent.futures import ThreadPoolExecutor
-from email import message_from_string
-from typing import List, Union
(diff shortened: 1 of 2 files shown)


deanq (approved):
lgtm

PR #421, opened 2025-06-04
fix: keep `allowed_cuda_versions` as string

deanq:
Backend replies back a string of comma-separated values. This PR keeps input and output symmetrical.


pandyamarut (approved):
/LGTM

PR #425, opened 2025-06-12
fix: ping fails when a process is blocking GIL

deanq:
Solution: run ping heartbeat in a separate Python process to its own GIL.

Tested by overriding an Endpoint's Docker Configuration > Container Start Command with:
```
/bin/bash -c "apt-get update && \
apt-get install -y git && \
pip install git+https://github.com/runpod/runpod-python@fix-unblock-ping --no-cache-dir && \
<insert original START CMD>"
```

### DIFF
--- a/runpod/serverless/modules/rp_ping.py
+++ b/runpod/serverless/modules/rp_ping.py
@@ -4,9 +4,8 @@ The heartbeat is responsible for sending periodic pings to the Runpod server.
 """
 
 import os
-import threading
 import time
-
+from multiprocessing import Process
 import requests
 from urllib3.util.retry import Retry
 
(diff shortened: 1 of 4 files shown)


runpod/serverless/modules/rp_ping.py
```diff
@@ -32,9 +31,10 @@ def __init__(self, pool_connections=10, retries=3) -> None:
         self.PING_URL = self.PING_URL.replace("$RUNPOD_POD_ID", WORKER_ID)
         self.PING_INTERVAL = int(os.environ.get("RUNPOD_PING_INTERVAL", 10000)) // 1000
 
+        # Create a new HTTP session
         self._session = SyncClientSession()
         self._session.headers.update(
-            {"Authorization": f"{os.environ.get('RUNPOD_AI_API_KEY')}"}
+            {"Authorization": os.environ.get("RUNPOD_AI_API_KEY", "")}
         )
```
justinwlin, line 38: Just wondering, is this an actual possible situation that can occur where there is no api_key?

Would it cause an issue? I couldn't really tell what the RUNPOD_AUTH key is being used for.

deanq: It was previously doing "None" which made no sense.

PR #451, opened 2025-08-04
feat: Implement pod restart API functionality

### LINKED ISSUE #450: Restart API not working
Bug: Runpod restart API not working

Endpoint: POST https://rest.runpod.io/v1/pods/{pod-id}/restart

Description:
When calling the restart API, it returns a success response (200 OK) but the pod does not actually restart. There is no change in pod state or behavior.

Steps to Reproduce:

Send a POST request to the above endpoint with a valid pod ID and API key

Observe the response

Check pod status — it remains unchanged

Expected Result:
The pod should stop and start again, similar to a manual restart from the dashboard.

Actual Result:
API returns success, but pod continues running without restarting.

Notes:

Tried with multiple pods and tools (curl, Postman)

Manual restart from dashboard works fine

kowshik24:
Fixes #450 - Add missing restart API for pods

- Add generate_pod_restart_mutation() to create GraphQL mutation
- Add restart_pod() high-level API function in ctl_commands.py
- Export restart_pod function in main module __init__.py
- Add comprehensive unit tests for both mutation and API function
- Update README.md with restart_pod usage example
- Update examples/graphql_wrapper.py with restart functionality

This implementation provides a proper restart API that preserves pod data unlike the stop/resume workaround which resets the pod state. The restart functionality allows users to restart pods without losing their data, addressing the core issue reported in #450.

The API follows the same pattern as other pod management functions (stop_pod, resume_pod, terminate_pod) and includes proper error handling and response parsing.

Resolves: #450


kowshik24:
@deanq could you please check this pull request about the restart api?

justinwlin:
Unfortunately this doesn't work. When trying to run it I get that there is no graphql mutation for this, and when looking deeper it does seem that there is other issues here that needs to be addressed first internally. Thanks for the PR!

I've filed a bug with the runpod team internally on the issue.

PR #453, opened 2025-08-21
Feat: Adding Ability to add API key per Request / Endpoint

justinwlin:
# Overview

Internal ticket AE-1090, resolution to add feature for client level / request level API key. 

Follow up:
Will need to update these docs
https://docs.runpod.io/sdks/python/endpoints

# GraphQL Calls

GraphQL calls I don't believe have a client / are stateless:
So the way those can work is such that we do:
```
gpus = runpod.get_gpu("NVIDIA A100 80GB PCIe", ...apikey....)
```

# Endpoint:

These have an internal state, so we can instantiate the Endpoint with the api key:
Before:
```
runpod.api_key = os.getenv("RUNPOD_API_KEY")

endpoint = runpod.Endpoint("YOUR_ENDPOINT_ID")
```

Now supported:
```
endpoint = runpod.Endpoint("YOUR_ENDPOINT_ID", API KEY)
```

### DIFF
--- a/runpod/endpoint/asyncio/asyncio_runner.py
+++ b/runpod/endpoint/asyncio/asyncio_runner.py
@@ -3,7 +3,7 @@
 # pylint: disable=too-few-public-methods,R0801
 
 import asyncio
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from runpod.endpoint.helpers import FINAL_STATES, is_completed
 from runpod.http_client import ClientSession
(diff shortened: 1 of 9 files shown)


justinwlin:
Make this a per client rather than per client / per request

runpod/endpoint/asyncio/asyncio_runner.py
```diff
@@ -112,22 +116,40 @@ async def cancel(self) -> dict:
...
+        self.api_key = api_key or global_api_key
+        
+        if self.api_key is None:
+            raise RuntimeError("API key must be provided or set globally")
+        
         self.headers = {
             "Content-Type": "application/json",
-            "Authorization": f"Bearer {api_key}",
+            "Authorization": f"Bearer {self.api_key}",
         }
```
justinwlin, line 148: Main thing here is that we want to figure out which API key to use:
```
        self.api_key = api_key or global_api_key
```

Save it in the header:
```
        self.headers = {
            "Content-Type": "application/json",
            "Authorization": f"Bearer {self.api_key}",
        }
```

Then whenever we run we use it:
```
        job_headers = self.headers.copy()
        job_headers["X-Request-ID"] = json_resp["id"]
        return Job(self.endpoint_id, json_resp["id"], self.session, job_headers)
```

deanq (approved):

PR #515, opened 2026-06-22
fix: concurrency correctness

hydrogen7797:


deanq (commented):
Review notes from a PR walkthrough. Two inline comments below.

runpod/endpoint/runner.py
```diff
@@ -44,8 +44,20 @@ def __init__(self, api_key: Optional[str] = None):
...
-        retries = Retry(total=5, backoff_factor=1, status_forcelist=[408, 429])
-        self.rp_session.mount("http://", HTTPAdapter(max_retries=retries))
+        # Only auto-retry idempotent methods. POST paths here (/run, /runsync)
+        # are non-idempotent and have no idempotency key, so a retry after a
+        # lost response or transient 5xx could create duplicate jobs.
+        retries = Retry(
+            total=5,
+            backoff_factor=1,
+            status_forcelist=[408, 429, 500, 502, 503, 504],
+            allowed_methods=frozenset(["GET"]),
```
deanq, line 54: Two things on this retry policy:

1. **Behavior change**: POSTs are no longer retried on 429. Previously all methods retried on 408/429; now `/run`, `/runsync`, `/cancel` retry on nothing. That's the right call for duplicate-job safety given there's no idempotency key, but a 429 is a rejected request that is safe to retry — worth documenting the tradeoff rather than leaving it implicit in the `allowed_methods` blanket.
2. **No test coverage**: this is the headline "don't create duplicate jobs" fix and nothing guards it. A future refactor could silently reintroduce POST retries. Please add a regression test asserting the mounted adapter's `Retry.allowed_methods == {"GET"}` and that both `http://` and `https://` are mounted.

runpod/endpoint/helpers.py
```diff
@@ -1,6 +1,6 @@
 """ Helper functions for the Runpod Endpoint API. """
 
-FINAL_STATES = ["COMPLETED", "FAILED", "TIMED_OUT"]
+FINAL_STATES = ["COMPLETED", "FAILED", "TIMED_OUT", "CANCELLED"]
```
deanq, line 3: Good catch aligning `FINAL_STATES` with `is_completed`. This is broader than a cleanup, though: `FINAL_STATES` is consumed directly in `runner.py` `stream()` (and the asyncio runner). Before this change, streaming a `CANCELLED` job never hit the `break` — `CANCELLED not in FINAL_STATES` stayed true, so the loop spun forever. Adding `CANCELLED` fixes that latent infinite loop. Please add a test that `stream()` terminates on `CANCELLED` (or at minimum assert `"CANCELLED" in FINAL_STATES`) so the fix is locked in.

PR #534, opened 2026-07-09
feat!: apps SDK, decorator-native GPU compute with live dev sessions and a new rp CLI

KAJdev:
Brings decorator-native GPU compute into the core `runpod` package, with queue, API, and task resources sharing one app lifecycle and a live development workflow.

Worker execution is versioned independently through `runpod-sdk-runtime`, allowing runtime images and SDK clients to evolve on separate release cycles. Flash lifecycle commands live under `rp flash`, while account, secret, registry, pod, and SSH commands stay at the root.

Completes CON-605
Completes CON-601
Completes CON-750


KAJdev:
@capy-ai review

runpod/apps/placement.py
```diff
@@ -0,0 +1,216 @@
...
+
+def candidates(spec, stock: StockMap) -> Set[str]:
+    """datacenters where a resource is schedulable.
+
+    hardware needs stock in the DC (any of the resource's acceptable
+    devices/flavors), intersected with an explicit datacenter pin.
+    """
+    allowed = {dc.value for dc in DataCenter.all()}
+    if spec.datacenter:
+        allowed &= {str(d) for d in spec.datacenter}
```
deanq, line 151: **Critical: datacenter enum pin makes the resource schedulable nowhere.**

`allowed &= {str(d) for d in spec.datacenter}` — `DataCenter` is a `(str, Enum)`, so `str(DataCenter.EU_RO_1)` is `"DataCenter.EU_RO_1"`, not `"EU-RO-1"`. Line 149 builds the base set with `dc.value`, so the intersection is always empty for an enum pin.

Reproduced on 3.14.6:
- `datacenter=DataCenter.EU_RO_1` → `PlacementError` ("schedulable nowhere") for a valid DC.
- `datacenter=[DataCenter.EU_RO_1]` → also corrupts the deploy manifest: `locations="DataCenter.EU_RO_1"`.

Plain strings (`"EU-RO-1"`) work, so this silently punishes the idiomatic enum API the SDK exports. Fix: intersect on `d.value` (or normalize via `DataCenter.from_string`), and normalize inputs in `_datacenter_list` (app.py:302-310) so case/underscore variants like `"us_ca_2"` also resolve.

runpod/apps/dev.py
```diff
@@ -0,0 +1,420 @@
...
+            resource = _resource_of(name)
+            emit(sink, "deleting", resource)
+            try:
+                await self.api.delete_endpoint(endpoint_id)
+                emit(sink, "deleted", resource)
+                log.info("deleted dev endpoint %s (%s)", name, endpoint_id)
+            except Exception as exc:
+                emit(sink, "delete_failed", resource)
+                log.warning("failed to delete dev endpoint %s: %s", endpoint_id, exc)
+        self._endpoints.clear()
```
deanq, line 410: **Critical: teardown orphans billed endpoints on delete failure.**

`self._endpoints.clear()` runs unconditionally after the loop, but the loop's `except` (lines 407-408) only warns. If `delete_endpoint` raises (expired auth, transient GraphQL/network error), the endpoint keeps running and billing, and its id is wiped from tracking — no handle to reclaim it later. This is the orphan-on-teardown class (SLS-344).

Collect failures (mirror `manage.py`'s `UndeployResult.failures`), remove only the entries that were actually deleted, and raise/return an aggregate so the caller and user learn the endpoints leaked.

runpod/apps/dev.py
```diff
@@ -0,0 +1,420 @@
...
+        desired: Dict[str, tuple] = {}
+        for app in apps:
+            for handle in self._provisionable(app):
+                name = dev_endpoint_name(app.name, handle.spec.name)
+                desired[name] = (app, handle)
+
+        # delete endpoints whose resources disappeared
+        for name in list(self._endpoints):
+            if name not in desired:
+                endpoint_id = self._endpoints.pop(name)
```
deanq, line 337: **Critical: reconcile pops tracking before the delete succeeds.**

`endpoint_id = self._endpoints.pop(name)` happens before the guarded `delete_endpoint`; on failure the `except` only warns (line ~344) and the id is already gone from `self._endpoints`, so no later teardown retries it → untracked, billing orphan. Pop only after a confirmed delete.

runpod/apps/targets.py
```diff
@@ -0,0 +1,886 @@
...
+        timeout: float = DEFAULT_TIMEOUT_SECONDS,
+    ) -> Dict[str, Any]:
+        return await _request_json(
+            method,
+            f"{self.base_url}/{path}",
+            self._headers(),
+            timeout,
+            payload=payload,
+            app_name=self._app_name,
+            resource_name=self._resource_name,
```
deanq, line 384: **Critical: TaskJob.wait() leaks the pod on timeout/error.**

The `finally` only calls `terminate()` when `self._done` is True. If `wait(timeout=…)` raises `TimeoutError` (or `poll_result()` raises), `_done` stays False and the pod runs — billed — until the 1h `terminateAfter` backstop. `cancel()` exists for exactly this teardown but is never invoked on the failure path. Terminate on any exit from `wait()`.

runpod/apps/targets.py
```diff
@@ -0,0 +1,886 @@
...
+    import time
+
+    if on_status is not None:
+        on_status(data)
+    deadline = time.monotonic() + timeout
+    interval = 0.5
+    # observed polls (dev sessions) stay tight so the worker id
+    # surfaces fast enough for log streams to attach in realtime
+    max_interval = 1.0 if on_status is not None else 5.0
+    while data.get("status") not in FINAL_STATUSES:
```
deanq, line 320: **Bug: a non-terminal response is returned as the successful result.**

Inside `while data.get("status") not in FINAL_STATUSES:`, when `data` has no `id` the code `return data`. `unwrap_job_output` then sees a non-FAILED, error-less envelope and returns it verbatim. Verified: `{"status": "IN_QUEUE"}` (or any partial/gateway body without an id) is handed back to the caller as the function's return value — a never-completed job masquerading as success. Treat "non-terminal + no job id" as an error instead of returning it.

runpod/apps/targets.py
```diff
@@ -0,0 +1,886 @@
...
+            merged["policy"] = {**merged["policy"], **value}
+        else:
+            merged[key] = value
+    return merged
+
+
+def unwrap_job_output(data: Dict[str, Any]) -> Any:
+    """extract output from a runsync-style response, raising on failure."""
+    if data.get("status") == "FAILED" or data.get("error"):
+        err = data.get("error") or data.get("output", {}).get("error", "unknown")
```
deanq, line 117: **Bug: unwrap_job_output raises AttributeError, masking the real failure.**

`data.get("output", {}).get("error", ...)` assumes `output` is a dict, but `LiveTarget`'s generator handler aggregates output into a list (~line 641). A failed live generator yields `{"status":"FAILED","output":[...]}` with no top-level `error`, so `[...].get("error")` raises `AttributeError` instead of a clean `RemoteExecutionError`. Guard with `isinstance(output, dict)` before `.get`.

runpod/apps/targets.py
```diff
@@ -0,0 +1,886 @@
...
+        live api endpoints materialize their routes from module source
+        pushed over /_runpod/sync (once per source change), keeping dev
+        behavior identical to deployed. the spec rides along so the
+        worker can install the resource's dependencies before @init.
+        """
+        self._source_target = target
+        self._source_resource = resource
+        self._source_spec = spec
+
+    async def _sync_source(self, timeout: float) -> None:
```
deanq, line 590: **Concurrency: unprotected shared state in _sync_source.**

`self._synced_hash` is read and mutated around an `await` with no `asyncio.Lock`. Two concurrent `.remote()`/`.request()` calls both observe the stale hash and both POST the full module source to `/_runpod/sync` — duplicate uploads, and racing in-flight source states can land the worker on the wrong revision. Guard the check-and-set with an `asyncio.Lock` (the project standard for async shared state).

runpod/apps/build.py
```diff
@@ -0,0 +1,375 @@
...
+    cmd.extend(requirements)
+
+    proc = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    count = 0
+    for line in proc.stdout:
```
deanq, line 287: **Bug: vendor() can deadlock draining pip's pipes sequentially.**

The code reads `proc.stdout` to exhaustion, then `proc.stderr.read()`. If pip writes more than the pipe buffer (~64KB) to stderr (resolver backtracking, deprecation spam) before stdout closes, the child blocks writing stderr while the parent blocks reading stdout → hang with no error. Use `proc.communicate()` (or drain both concurrently).

runpod/apps/spec.py
```diff
@@ -0,0 +1,289 @@
...
+
+@dataclass
+class ResourceSpec:
+    """declarative config for one app resource."""
+
+    kind: ResourceKind
+    name: str
+    gpu: Optional[List[str]] = None
+    cpu: Optional[List[str]] = None
+    gpu_count: int = 1
```
deanq, line 156: **Missing validation: gpu_count is unchecked.**

`gpu_count: int = 1` is declared but `__post_init__` validates every other numeric field and skips this one, so `gpu_count=0` / `-2` flows straight into the manifest (spec.py:233) and pod input (tasks.py:137). Add a `>= 1` guard in `__post_init__`, plus a test.

runpod/apps/spec.py
```diff
@@ -0,0 +1,289 @@
...
+    normalized = str(version).strip()
+    if normalized not in CUDA_VERSIONS:
+        raise InvalidResourceError(
+            f"min_cuda_version must be one of "
+            f"{', '.join(sorted(CUDA_VERSIONS))}, got {version!r}"
+        )
+    return normalized
+
+
+@dataclass
```
deanq, line 148: **Type design: ResourceSpec is a mutable public type whose validation is bypassable.**

`ResourceSpec` is exported in `__all__` but its `normalize_*` validation lives in the `App.queue/task/api` decorators, not `__post_init__`. Direct construction — `ResourceSpec(kind=QUEUE, name="x", scaler_type="BOGUS", workers=(9,1), min_cuda_version="99")` — constructs cleanly, bypassing every guard. Handles also mutate it after construction (`spec.schedule=…`, `spec.routes=…`), so post-construction invalid states reach `to_manifest()`. For a permanent public SDK surface, make it `frozen=True, slots=True` and move validation into the construction path (have handles build a new spec rather than mutate).

runpod/apps/stubs.py
```diff
@@ -0,0 +1,150 @@
...
+        # id-based stubs still route through the sentinel; the id is
+        # passed as the resource key and resolved server-side
+        resource = self.resource_name or self.resource_id or ""
+        return SentinelTarget(self.app_name, self.env, resource)
+
+
+class Queue(_StubBase):
+    """client for a queue resource deployed elsewhere."""
+
+    def __init__(self, **kwargs: Any):
```
deanq, line 52: **Type design: Queue/Api constructors erase their public signature.**

Public constructors are typed `def __init__(self, **kwargs: Any)`; the real params (`app`, `name`, `id`, `env`) live only on `_StubBase`. `runpod.Queue(app="a", nmae="transcribe")` passes type-checking and forwards the typo into `**kwargs`. On a stable public API this is the worst encapsulation defect in the surface — use explicit typed parameters, and express the "exactly one of name/id" rule via `Queue.by_name(...)` / `Queue.by_id(...)` classmethods rather than a runtime XOR.

runpod/apps/datacenter.py
```diff
@@ -0,0 +1,58 @@
...
+    @classmethod
+    def from_string(cls, value: str) -> "DataCenter":
+        """parse a datacenter id, accepting lowercase and underscores."""
+        normalized = value.strip().upper().replace("_", "-")
+
+        try:
+            return cls(normalized)
+        except ValueError:
+            valid = ", ".join(dc.value for dc in cls)
+            raise ValueError(
```
deanq, line 35: **Exception chaining: re-raise without `from`.**

`except ValueError: raise ValueError(...)` drops the cause and trips Ruff B904 (CI) — and every invalid datacenter string hits it. Use `raise ValueError(...) from None` (intentional) or `from exc`.

runpod/rp_cli/main.py
```diff
@@ -0,0 +1,1031 @@
...
+        asyncio.run(_run())
+    except KeyboardInterrupt:
+        # ctrl-c ends the log follow; not an error
+        pass
+
+
+# ----------------------------------------------------------------- login
+
+
+@cli.command()
```
deanq, line 525: **Compat regression: `rp login` drops multi-profile support.**

The old `runpod config` accepted `--profile` and wrote credentials for any named profile; `login` has no `--profile` and calls `set_credentials(api_key, overwrite=True)`, which hardcodes `profile="default"` (config/functions.py:18). A user who ran `runpod config --profile staging <key>` has no CLI path under the new command, and `rp login` silently overwrites their `default` profile. Add `--profile` (or document the removal in the migration notes).

deanq (changes requested):
Review from a facade-delegation + correctness pass on the apps SDK (I'm consuming this contract from runpod-flash's compat facade).

**Requesting changes** on the two verified merge-blockers below; the rest are important-but-negotiable. Three findings were reproduced locally (noted inline).

Critical:
- `placement.py` datacenter enum pin -> schedulable nowhere (verified 3.14.6)
- dev/task teardown orphans billed endpoints & pods on the failure path (the SLS-344 class)

Facade note: the sentinel dispatch wire contract matches cleanly and delegates today. Three seams still force flash to keep its own code and are worth landing here: a **public** header builder (facade currently imports the private `targets._headers`), an id-only stub/client (Queue/Api require `app=`), and a remote-class execution seam (apps is function-only; flash's RemoteClassWrapper has no landing spot). Happy to file these as separate issues if you'd prefer.

KAJdev:
Should be ready for a re-review. Took a look at all these comments, main things were addressed.

- rebase with the new REST wrapper API client for all supported operations
 - resource validation and GPU/datacenter/volume placement, including multi-GPU and multi-volume configurations
 - endpoint and pod cleanup across partial startup, failures, timeouts, and cancellation plus retry policies that avoid duplicating mutations
 - nested sync/async worker calls, concurrent source synchronization, result/error handling, and dependency packaging
 - independently versioned runtime images, custom-image bootstrapping, and runtime publication.
 - .remote() calls on queue based endpoints now use /runsync first, then switch to polling during longer requests

runpod/apps/tasks.py
```diff
@@ -0,0 +1,406 @@
...
+    from .gpu import POOLS_TO_TYPES, GpuGroup
+
+    if not gpu:
+        return sorted(
+            {t.value for types in POOLS_TO_TYPES.values() for t in types}
+        )
+    names: List[str] = []
+    for entry in gpu:
+        try:
+            names.extend(GpuGroup(entry).device_names() or [entry])
```
deanq, line 75: `gpu="any"` (and `GpuGroup.ANY`) can't deploy. `spec.gpu` normalizes to `["any"]`, then `_device_names(["any"])` → `GpuGroup("any").device_names()` is `[]`, so `[] or ["any"]` falls back to the literal `["any"]` and `gpuTypeIdList=['any']`. The pod deploy API can't resolve `'any'` as a device name, so the task never launches — whereas `gpu=None` correctly expands to the full device list. Worth normalizing the ANY sentinel to the same full-list expansion as `None`. (Same enum→wire-value class as the `placement.py` datacenter fix from the last round, one module over.)

runpod/api/rest.py
```diff
@@ -87,8 +86,44 @@ def run_rest_request(
...
     return response.json()
+
+
+async def run_rest_request_async(
+    method: str,
+    path: str,
+    *,
+    api_key: Optional[str] = None,
+    params: Optional[Mapping[str, Any]] = None,
+    json: Optional[Mapping[str, Any]] = None,
```
deanq, line 103: `session._retry_connection = False` doesn't do anything — `aiohttp.ClientSession` has no `_retry_connection` attribute, so this just sets an unused instance attr. The comment says it prevents aiohttp from replaying PUT/DELETE after an ambiguous disconnect, but aiohttp doesn't auto-retry requests, so there's nothing to disable here. Either drop it as dead code, or — if replay of non-idempotent calls (`delete_endpoint`, `terminate_pod`) is a genuine concern — handle it explicitly.

deanq (commented):
Re-review after the REST-wrapper rebase — the earlier blockers all look resolved (placement datacenter enum, dev teardown orphaning, the targets.py pod leaks, etc.). A few new issues in the new/rebased code, inline below.

The `tasks.py` `gpu="any"` one is the same enum→device-name class as the `placement.py` fix from last round and would block task deploys; the `rp_info.py` `get_pod` change is a retry regression on the `rp pod connect` path. The rest are robustness/correctness.

runpod/cli/utils/rp_info.py
```diff
@@ -18,7 +18,9 @@ def get_pod_ssh_ip_port(pod_id, timeout=300):
     status = None
 
     while time.time() - start_time < timeout and (pod_ip is None or pod_port is None):
-        pod = get_pod(pod_id) or {}
+        pod = get_pod(pod_id)
```
deanq, line 21: This changes failure-path behavior: previously `pod = get_pod(pod_id) or {}` tolerated a transient `None`/empty response and the `while` loop retried until `timeout`. Now `if not pod: raise ValueError(...)` aborts on the first falsy return, so a pod that's still provisioning (or a momentary empty GraphQL response) makes `get_pod_ssh_ip_port` — and thus `rp pod connect` / `SSHConnection` — crash immediately instead of waiting for RUNNING. Suggest keeping the tolerate-and-retry within the timeout window, and only raising once the deadline is hit.

runpod/apps/gpu.py
```diff
@@ -0,0 +1,225 @@
...
+    def all(cls) -> List["GpuType"]:
+        """all gpu types except ANY."""
+        return [g for g in cls if g is not cls.ANY]
+
+
+POOLS_TO_TYPES = {
+    GpuGroup.ADA_24: [GpuType.NVIDIA_GEFORCE_RTX_4090],
+    GpuGroup.ADA_32_PRO: [GpuType.NVIDIA_GEFORCE_RTX_5090],
+    GpuGroup.ADA_48_PRO: [GpuType.NVIDIA_RTX_6000_ADA_GENERATION],
+    GpuGroup.ADA_80_PRO: [GpuType.NVIDIA_H100_80GB_HBM3],
```
deanq, line 104: The docstring (line 26) lists `ADA_80_PRO` as covering H100 PCIe + 80GB HBM3 + NVL, but `POOLS_TO_TYPES[ADA_80_PRO]` only has `NVIDIA_H100_80GB_HBM3` (PCIe and NVL aren't in `GpuType` at all). So `device_names()` omits the PCIe/NVL variants — a task requesting `ADA_80_PRO` can't schedule onto those machines, and `resolve_gpu_string("H100")` only ever matches the one HBM3 device. Either add the missing device types or narrow the docstring to match reality.

runpod/apps/api.py
```diff
@@ -0,0 +1,583 @@
...
+        if body is not None and endpoint_id:
+            current = await run_rest_request_async(
+                "GET",
+                f"/v2/serverless/{_path_segment(endpoint_id)}",
+                api_key=self._api_key,
+            )
+            requested_type = endpoint_input.get("type")
+            expected = (
+                ("LOAD_BALANCER" if requested_type == "LB" else "QUEUE")
+                if requested_type is not None
```
deanq, line 210: `run_rest_request_async` returns `None` on a 204 / empty body, but here `current = await run_rest_request_async('GET', ...)` is dereferenced right after with `current.get('type')`. If the GET yields no body, `current` is `None` and this raises `AttributeError` instead of a meaningful error. A `None` guard (treat as "no existing endpoint") would be safer.

runpod/apps/serialization.py
```diff
@@ -0,0 +1,101 @@
...
+    # last resort (repl): the bare function body with decorators
+    # stripped, since their context cannot travel
+    return _bare_function_source(fn)
+
+
+def _bare_function_source(fn: Callable) -> str:
+    """a function's own def, decorators removed."""
+    source = textwrap.dedent(inspect.getsource(fn))
+    tree = ast.parse(source)
+    for node in ast.walk(tree):
```
deanq, line 94: `ast.walk` yields nested defs too, so `_bare_function_source` picking the first `FunctionDef` whose name matches can select an inner function (or a same-named sibling in the parsed snippet) rather than the target — and `lines[node.lineno-1:]` then ships the wrong body to the worker. It's the REPL last-resort path, but it fails silently (wrong remote code) rather than erroring. Consider matching only the top-level / outermost def.

runpod/apps/api.py
```diff
@@ -0,0 +1,583 @@
...
+            timeout=10,
+        )
+
+    async def deploy_task_pod(
+        self, pod_input: Dict[str, Any], *, is_cpu: bool
+    ) -> Dict[str, Any]:
+        """provision with a termination deadline and public-ip requirement."""
+        pod_input = dict(pod_input)
+        if is_cpu:
+            instance_ids = pod_input.pop("instanceIds", None) or [
```
deanq, line 276: `instance_ids = pod_input.pop('instanceIds', None) or [pod_input.get('instanceId')]` becomes `[None]` when the input has neither key, and the loop then deploys a candidate with `instanceId=None` — producing a confusing server-side error instead of a clear client-side validation failure. Guard the empty case before the loop.

deanq:
hey @KAJdev — poking this one. the 7 threads from my last pass are still open and the code hasn't moved. couple of them are actual bugs, not nits:

- `apps/tasks.py` — `gpu="any"` / `GpuGroup.ANY` collapses to a literal `gpuTypeIdList=['any']`, so that path just can't deploy
- `apps/api.py` — `current.get('type')` on a maybe-`None` 204 body, plus the `instanceIds` fallback ending up `[None]` and deploying with `instanceId=None`
- `api/rest.py` — `session._retry_connection = False` does nothing (aiohttp has no such attr), so PUT/DELETE replay isn't actually prevented
- `cli/utils/rp_info.py` — the `raise ValueError` makes it bail on the first null response instead of retrying till timeout

the `gpu.py` docstring mismatch and the `serialization.py` nested-def thing are the minor two. ping me once these have a look and i'll re-review

runpod/api/rest.py
deanq: Re-opening — this one got resolved but the code is unchanged. `aiohttp.ClientSession` has no `_retry_connection` attribute (we pin `aiohttp >= 3.14.3`), so `session._retry_connection = False` just sets an inert instance attr that aiohttp never reads. PUT/DELETE replay isn't actually prevented and the comment above it is misleading. We already depend on `aiohttp-retry` — that's probably where this belongs. Can you either wire it up for real or drop the line + comment?

runpod/apps/serialization.py
deanq: Dropping this one on a closer look — `ast.walk` is breadth-first so the top-level def is always yielded before any nested same-named def, and `inspect.getsource(fn)` only returns `fn`'s own source so there's no sibling to collide with. Safe as written.

PR #570, opened 2026-08-19
feat: SLS-497 Add explicit prestart hooks to the runpod-python SDK

jasonwang-runpod:
## Problem

Workers often load models or start inference engines before calling `runpod.serverless.start()`. The SDK cannot observe that prestart work, so failures leave requests in `IN_QUEUE` status while useful info remains in worker logs.

## Solution

Add explicit prestart hooks that the SDK supervises before handler execution.

- Workers can register sync or async hooks with `@runpod.serverless.register_prestart_hook` and provide an optional timeout for the prestart phase.
- Queue-based workers take jobs concurrently with prestart, but don't run the handler function until prestart succeeds.
- A failure or timeout is attached to a request as `prestart_failed`. Then the worker drains and exits.
- Local input and single-worker hosted API mode run prestart before invoking or serving the handler. Realtime and multi-process hosted API modes are unsupported.
- Failure payloads can carry the last 16 KB of the worker's stdout and stderr. `RUNPOD_LOG_CAPTURE` gates this: `auto` (default) captures only when hooks are registered, `all` always captures, `off` never does. Note: log capture doesn't see child processes.
- A worker that doesn't register any hooks is unaffected.

## Testing

`uv run pytest -q`: 710 passed, 94.7% coverage.

An sglang endpoint was tested w/ different scenarios:
1) no hooks: COMPLETED, output only, no logs field
2) hook success: COMPLETED, engine boot ran inside the prestart hook
3) failing hook: FAILED (as expected) with reason `prestart_failed`
4) `prestart_timeout=5` with slow hook: FAILED (as expected) in 4.9s with `PrestartTimeout`; logs had the captured stdout
5) hook ok + handler raises, `capture=auto`: FAILED (as expected), with no logs field

Supersedes #567.


jasonwang-runpod:
bugbot run

tests/test_serverless/test_prestart.py
```diff
@@ -0,0 +1,267 @@
...
+
+            async def blocked_hook():
+                started.set()
+                await asyncio.Event().wait()
+
+            task = asyncio.create_task(run_prestart_hooks_async((blocked_hook,)))
+            await started.wait()
+            task.cancel()
+            with self.assertRaises(asyncio.CancelledError):
+                await task
```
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: the await is required for `assertRaises` to observe `CancelledError`.

runpod/serverless/modules/rp_prestart.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+async def _invoke_hook(hook: Callable[[], Any]) -> None:
+    if inspect.iscoroutinefunction(hook) or inspect.iscoroutinefunction(hook.__call__):
+        result = hook()
+    else:
+        result = await _run_sync_in_daemon(hook)
+
+    if inspect.isawaitable(result):
+        await result
```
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: `await result` executes and waits for the returned awaitable; it is not ineffective.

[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: `await result` executes and waits for the returned awaitable; it is not ineffective.

runpod/serverless/modules/rp_scale.py
```diff
@@ -136,17 +155,52 @@
...
+                if prestart_task in done:
+                    try:
+                        await prestart_task
+                    except BaseException:
+                        # Normal prestart failures are handled inside _run_prestart.
+                        self.kill_worker()
+                        raise
+
+                # Prestart completion does not end the worker; request loops do.
+                await request_loops_future
```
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: this await propagates request-loop failures; cleanup deliberately suppresses them.

runpod/serverless/modules/rp_prestart.py
```diff
@@ -0,0 +1,188 @@
...
+        await asyncio.shield(hook_task)
+    except asyncio.CancelledError as exc:
+        if hook_task.cancelled():
+            raise PrestartError(exc, _hook_name(hook)) from exc
+
+        # The caller cancelled the prestart task.
+        # Now cancel and drain the child task before propagating shutdown.
+        hook_task.cancel()
+        with contextlib.suppress(BaseException):
+            await hook_task
```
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: this await drains the cancelled child task before cancellation propagates.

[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: this await drains the cancelled child task before cancellation propagates.

tests/test_serverless/test_prestart.py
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: the await is required for `assertRaises` to observe `CancelledError`.

runpod/serverless/modules/rp_scale.py
```diff
@@ -136,17 +155,52 @@ def handle_shutdown(self, signum, frame):
...
```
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: this await propagates request-loop failures; cleanup deliberately suppresses them.

runpod/serverless/modules/rp_prestart.py
```diff
@@ -0,0 +1,188 @@
...
+    loop = asyncio.get_running_loop()
+    done = asyncio.Event()
+    results: list[Any] = []
+    errors: list[BaseException] = []
+    ctx = contextvars.copy_context()
+
+    def worker():
+        try:
+            results.append(ctx.run(hook))
+        except BaseException as exc:  # noqa: BLE001 - transferred to the event loop
```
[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: `BaseException` intentionally transfers control exceptions from the daemon thread to the event loop.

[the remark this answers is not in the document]
jasonwang-runpod: Rejecting: `BaseException` intentionally transfers control exceptions from the daemon thread to the event loop.

jhcipar (approved):

runpod/serverless/modules/rp_capture.py
```diff
@@ -0,0 +1,143 @@
...
+    """Install the tee proxy on stdout/stderr, if this worker wants capture.
+
+    Idempotent. `hooks_registered` is what `auto` mode keys off, and the caller
+    passes it in so this module stays independent of the prestart registry.
+    """
+    mode = capture_mode()
+    if mode == CAPTURE_OFF:
+        return
+
+    if mode == CAPTURE_AUTO and not hooks_registered:
```
deanq, line 104: Registering any prestart hook flips `hooks_registered=True`, which in the default `auto` mode enables the stdout/stderr tee for the whole worker. Since `run_job` already wraps every handler in `capture()`, a handler that prints sensitive data and then raises will have those lines captured into `error_info['logs']` and returned to the caller in the job result.

Concretely: a worker adopts an unrelated prestart hook -> capture turns on -> a handler does `print(f"HF_TOKEN={token}")` then raises -> the token lands in the error payload polled by the client.

The coupling (prestart-hook presence silently enables handler-failure log capture) is non-obvious and has data-exfiltration impact. Suggest decoupling: gate handler-failure capture on an explicit opt-in rather than on hook registration, or redact/scrub captured logs before embedding them in the returned error.

runpod/serverless/modules/rp_scale.py
```diff
@@ -338,14 +432,28 @@ async def stop_job(self, job_id: str) -> bool:
...
         """
         Process an individual job. This function is run concurrently for multiple jobs.
         """
         batch_id_token = _set_batch_id(job.get("batchId"))
         try:
             log.debug("Handling Job", job["id"])
 
+            # Hold the handler until every registered prestart hook finishes.
+            if self.prestart_hooks:
+                if not await self._wait_for_prestart():
```
deanq: If shutdown begins while prestart is still running, `_wait_for_prestart()` returns False, and `handle_job` logs "leaving this request for another worker" and returns -- without calling `send_result` or re-queuing the job.

But this worker already claimed the job. If the platform does not re-dispatch an already-assigned request, it's now neither completed nor failed: the caller waits out its queue TTL with no result -- a silent failure.

Suggest failing the claimed request explicitly on this path (send an error result) rather than relying on server-side re-dispatch that may never happen.

runpod/serverless/modules/rp_scale.py
```diff
@@ -369,3 +477,37 @@ async def handle_job(self, session: ClientSession, job: dict):
...
+
+    async def _run_prestart(self):
+        """Run hooks beside queue intake, then open the handler gate."""
+        try:
+            self._prestart_error = await run_prestart_phase(
+                self.prestart_hooks, self.config.get("prestart_timeout")
+            )
+        finally:
+            # Always release held handlers.
+            self._prestart_ready.set()
```
deanq, line 553: `_run_prestart` sets `_prestart_ready` in a `finally` block. If prestart is cancelled during shutdown (mid-hook), `run_prestart_phase` raises CancelledError, which still propagates through `finally: self._prestart_ready.set()` -- so `_prestart_ready` becomes set while `_prestart_error` remains None.

A handler concurrently in `_wait_for_prestart` then observes `_prestart_ready.is_set() == True`, returns True, sees `_prestart_error is None`, and runs the handler against a half-initialized environment (e.g. `model` still None) even though prestart never actually completed.

Suggest distinguishing "prestart finished" from "prestart cancelled" -- e.g. set an explicit error/cancelled state before setting the ready event, so `_wait_for_prestart` doesn't release the handler on cancellation.

runpod/serverless/modules/rp_scale.py
```diff
@@ -136,17 +148,52 @@ def handle_shutdown(self, signum, frame):
...
+                if prestart_task in done:
+                    try:
+                        await prestart_task
+                    except BaseException:
+                        # Normal prestart failures are handled inside _run_prestart.
+                        self.kill_worker()
+                        raise
+
+                # Prestart completion does not end the worker; request loops do.
+                await request_loops_future
```
deanq: On prestart failure the worker can't actually terminate until in-flight long-polls return. `kill_worker()` only sets an event the loops check at their top, but `run()` blocks here on `await request_loops_future` -- and `get_jobs` / `monitor_stop_signals` are parked mid-`wait_for` with 90s fetch timeouts. So `_terminate_unhealthy(1)` (line 196) is deferred up to ~90s after the failure was already reported, leaving a known-broken worker alive and schedulable.

This partially defeats the point of `rp_fitness._terminate_unhealthy`, whose whole rationale is os._exit precisely because "sys.exit can hang forever and the worker keeps serving jobs." Routing the prestart-failure exit through the graceful loop-drain reintroduces exactly that hang.

Suggest calling the hard-exit path (or cancelling the request loops) immediately on prestart failure, rather than awaiting `request_loops_future` first.

runpod/serverless/modules/rp_prestart.py
```diff
@@ -0,0 +1,225 @@
...
+        await asyncio.shield(hook_task)
+    except asyncio.CancelledError as exc:
+        if hook_task.cancelled():
+            raise PrestartError(exc, _hook_name(hook)) from exc
+
+        # The caller cancelled the prestart task.
+        # Now cancel and drain the child task before propagating shutdown.
+        hook_task.cancel()
+        with contextlib.suppress(BaseException):
+            await hook_task
```
deanq, line 201: In the cancellation branch, after `hook_task.cancel()` this does `with contextlib.suppress(BaseException): await hook_task`. If an async hook catches CancelledError and keeps running (e.g. a retry loop that ignores cancellation), `await hook_task` never returns -- so the prestart-timeout / shutdown path that got us here never completes, and the worker deadlocks instead of terminating.

Suggest bounding this await (e.g. `await asyncio.wait_for(hook_task, timeout=...)` and proceeding regardless on timeout) so an uncooperative hook can't wedge the timeout/shutdown path.

runpod/serverless/modules/rp_scale.py
```diff
@@ -60,6 +69,7 @@ def __init__(self, config: Dict[str, Any]):
         self.concurrency_modifier = _default_concurrency_modifier
         self.jobs_fetcher = get_job
         self.jobs_fetcher_timeout = 90
+        self.prestart_claim_timeout = 10
```
deanq: `self.prestart_claim_timeout = 10` is a bare inline magic number with no name explaining intent and no unit. This PR already defines named constants elsewhere (e.g. `MAX_CAPTURED_CHARS` in rp_capture.py), so this is inconsistent. Suggest a module-level named constant with the unit in the name, e.g. `PRESTART_CLAIM_TIMEOUT_SECONDS = 10`.

deanq (commented):
Reviewed the prestart-hooks + capture changes. The concurrency design (concurrent intake with a gated handler, failing the claimed request on startup failure) is sound and well-tested. Flagging a few correctness/security gaps on the failure and shutdown paths, plus one nit.

One cross-cutting note: the prestart-failure path routes termination through `rp_fitness._terminate_unhealthy` (the fitness force-kill helper) but defers its `os._exit` behind the graceful loop-drain -- see the inline note on `run()`. That partially undoes the hard-exit guarantee fitness relies on. Details inline.

deanq (changes requested):

runpod/serverless/modules/rp_capture.py
deanq: Addressed by the `handler_capture_enabled()` split — `run_job`/`run_job_generator` now capture only when `RUNPOD_LOG_CAPTURE` is `all`, and `capture()` no-ops the ring buffer when disabled. In `auto`, handler output is no longer captured just because a prestart hook happens to be registered. Resolving.

runpod/serverless/modules/rp_scale.py
deanq: Fixed — `handle_job` no longer returns silently when shutdown begins before prestart is ready. It now fails the held request with the new `prestart_cancelled` payload (`build_prestart_cancelled_payload`) instead of dropping it for another worker. Resolving.

runpod/serverless/modules/rp_scale.py
deanq: Resolved by the new `_prestart_succeeded` flag. The `finally` still wakes held handlers, but they now branch on the flag rather than treating "ready" as success — so a prestart cancelled mid-hook is reported as `prestart_cancelled` instead of being mistaken for a successful start. Resolving.

runpod/serverless/modules/rp_scale.py
deanq: Fixed — the busy-wait on `current_occupancy()` is gone. On prestart failure, `run()` now cancels the stop-signal and job-take loops and bounds failure reporting with `prestart_failure_report_timeout` (15s, well under the shared 600s HTTP timeout), so worker termination no longer blocks on an in-flight long-poll. Resolving.

runpod/serverless/modules/rp_prestart.py
deanq: Addressed — after `hook_task.cancel()` the drain is now bounded by `PRESTART_HOOK_CANCEL_GRACE_SECONDS` (1s). A hook that swallows `CancelledError` and keeps running raises `PrestartCancellationTimeout` instead of hanging shutdown indefinitely. Resolving.

runpod/serverless/modules/rp_scale.py
deanq: Fixed — extracted to the named module constant `PRESTART_CLAIM_TIMEOUT_SECONDS` (with `PRESTART_FAILURE_REPORT_TIMEOUT_SECONDS` alongside it), each with a comment on intent and unit. Resolving.

deanq (approved):
All six review threads addressed and resolved; CI green across 3.10–3.14, e2e, and CodeQL. The prestart failure/cancellation paths now fail held requests explicitly (prestart_cancelled), bound the hook-cancel drain, and cap failure reporting below the HTTP timeout. LGTM.

PR #578, opened 2026-08-26
feat(serverless): coordinate early health checks once per container

justinwlin:
Worker health checks previously waited until `serverless.start()`, often after expensive model loading. This change runs the built-in hardware checks (RAM, disk, CUDA version, native GPU test) at the first `import runpod` in a Serverless worker, and keeps network, Python CUDA initialization, GPU compute, and customer checks at worker start.

**Eligibility.** Early checks require both `RUNPOD_ENDPOINT_ID` and `RUNPOD_WEBHOOK_GET_JOB`, and skip `RUNPOD_TEST`, `--test_input`, and `--rp_serve_api` invocations. The trigger lives in the top-level `runpod/__init__.py`, so it does not depend on eager `serverless` import.

**Once per process tree.** A process whose early pass passes sets `RUNPOD_EARLY_FITNESS_CHECKS_DONE=1`. Child processes inherit it (multiprocessing spawn re-importing the handler, subprocesses, shell wrappers) and skip the pass instead of repeating the GPU probes. A failed check exits before the marker is set. No lock files, procfs identity, or shared state.

**Flags.** `RUNPOD_DEFER_FITNESS_CHECKS=true` restores worker-start timing. `RUNPOD_SKIP_FITNESS_CHECKS=true` disables all checks. Late threshold changes are applied at worker start and only rerun the affected checks.

**Also in this PR.** Network check is deferred to worker start, targets the worker API host, and retries within one bounded budget. Registration is atomic and setup failures at worker start go through the same force-exit path. Logger redaction fixes. Module moves to `runpod/_health/` with import aliases at the old paths.

**Not in this PR.** Realtime workers keep their `main` behavior (no fitness checks). `feat/apps-sdk` will need the one-line `run_import_checks()` call carried into its lazy `__init__` when main is merged in.

Validation: full suite 694 passed, 93% coverage. A real `multiprocessing` spawn child test confirms the child inherits the marker and runs zero probes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/runpod/serverless/__init__.py
+++ b/runpod/serverless/__init__.py
@@ -16,7 +16,7 @@ from ..version import __version__ as runpod_version
 from . import worker
 from .modules.rp_logger import RunPodLogger
 from .modules.rp_progress import progress_update
-from .modules.rp_fitness import register_fitness_check
+from .modules.rp_fitness import register_fitness_check, run_startup_fitness_checks
 from .utils.rp_volume_cache import VolumeCache
 
 __all__ = [
@@ -28,6 +28,11 @@ __all__ = [
 ]
 
 log = RunPodLogger()
+
+# Validate the worker environment now, at import, rather than waiting for
+# start() -- which a handler module only reaches after it has loaded its model.
+# No-op outside a real worker; see run_startup_fitness_checks.
+run_startup_fitness_checks()
 
 
 # ---------------------------------------------------------------------------- #
--- a/runpod/serverless/__init__.py
+++ b/runpod/serverless/__init__.py
@@ -29,9 +29,8 @@ __all__ = [
 
 log = RunPodLogger()
 
-# Validate the worker environment now, at import, rather than waiting for
-# start() -- which a handler module only reaches after it has loaded its model.
-# No-op outside a real worker; see run_startup_fitness_checks.
+# Check the environment here rather than in start(), which a handler module
+# only reaches after loading its model. No-op outside a real worker.
 run_startup_fitness_checks()
 
 
--- a/runpod/serverless/modules/rp_fitness.py
+++ b/runpod/serverless/modules/rp_fitness.py
@@ -67,6 +67,23 @@ def _env_flag(name: str) -> bool:
     return os.environ.get(name, "").strip().lower() in ("1", "true", "yes", "on")
 
 
+def defer_to_worker_start(func: Callable) -> Callable:
+    """
+    Mark a check as unsafe to run at import.
+
+    The import-time pass skips these; they run in run_worker as before. Used
+    for checks that initialize CUDA in this process -- doing that before the
+    handler module runs would leave a CUDA context in a process the handler
+    may later fork (vLLM, DeepSpeed), which CUDA does not support.
+    """
+    func._runpod_defer_to_worker_start = True
+    return func
+
+
+def _is_deferred(func: Callable) -> bool:
+    return getattr(func, "_runpod_defer_to_worker_start", False)
+
+
 def register_fitness_check(func: Callable) -> Callable:
     """
     Decorator to register a fitness check function.
@@ -222,7 +239,7 @@ def _ensure_system_checks_registered() -> None:
         log.debug("System fitness check module not found, skipping auto-registration")
 
 
-async def run_fitness_checks() -> None:
+async def run_fitness_checks(include_deferred: bool = True) -> None:
     """
     Execute all registered fitness checks sequentially at startup.
 
@@ -266,6 +283,9 @@ async def run_fitness_checks() -> None:
     _ensure_system_checks_registered()
 
     pending = [check for check in _fitness_checks if check not in _completed_checks]
+
+    if not include_deferred:
+        pending = [check for check in pending if not _is_deferred(check)]
 
     if not pending:
         log.debug("No pending fitness checks, skipping.")
@@ -334,7 +354,8 @@ def run_startup_fitness_checks() -> None:
     Run the built-in fitness checks at import, before the handler loads a model.
 
     A user's @register_fitness_check functions are registered after this import,
-    so they still run in run_worker, which skips whatever passed here.
+    so they still run in run_worker, which skips whatever passed here. Checks
+    marked with @defer_to_worker_start are also left to run_worker.
 
     No-ops outside a real worker (no RUNPOD_WEBHOOK_GET_JOB), when the checks
     are disabled or deferred, or inside a running event loop. Never raises: a
@@ -352,6 +373,6 @@ def run_startup_fitness_checks() -> None:
         return
 
     try:
-        asyncio.run(run_fitness_checks())
+        asyncio.run(run_fitness_checks(include_deferred=False))
     except Exception as exc:  # pragma: no cover - defensive
         log.warn(f"Startup fitness checks could not run: {exc}")
--- a/runpod/serverless/modules/rp_fitness.py
+++ b/runpod/serverless/modules/rp_fitness.py
@@ -51,29 +51,20 @@ def _terminate_unhealthy(code: int = 1) -> None:
 # Global registry for fitness check functions, preserves registration order
 _fitness_checks: list[Callable] = []
 
-# Checks that have already executed successfully. Fitness checks now run at
-# import time (see run_startup_fitness_checks) as well as in run_worker, so the
-# second pass must only execute checks registered after the first pass -- the
-# user's own @register_fitness_check functions, which are registered between
-# the two.
+# Checks that already passed. Checks run twice per worker -- at import and in
+# run_worker -- so the second pass only runs what was registered in between.
 _completed_checks: list[Callable] = []
 
-# Env var that disables every fitness check, built-in and user-registered.
+# Disables every check, built-in and user-registered.
 SKIP_FITNESS_CHECKS_ENV = "RUNPOD_SKIP_FITNESS_CHECKS"
 
-# Env var that restores the old behavior: checks run only when the worker
-# starts (after the handler module has loaded its model), not at import.
+# Keeps the checks but runs them only in run_worker, as before.
 DEFER_FITNESS_CHECKS_ENV = "RUNPOD_DEFER_FITNESS_CHECKS"
 
 
 def _env_flag(name: str) -> bool:
     """True if the env var is set to a truthy value."""
     return os.environ.get(name, "").strip().lower() in ("1", "true", "yes", "on")
-
-
-def fitness_checks_disabled() -> bool:
-    """True if the user has opted out of fitness checks entirely."""
-    return _env_flag(SKIP_FITNESS_CHECKS_ENV)
 
 
 def register_fitness_check(func: Callable) -> Callable:
@@ -263,10 +254,8 @@ async def run_fitness_checks() -> None:
         A failing check terminates the process via os._exit(1); this function
         does not return in that case and does not raise SystemExit.
     """
-    if fitness_checks_disabled():
-        log.info(
-            f"Fitness checks disabled via {SKIP_FITNESS_CHECKS_ENV}, skipping."
-        )
+    if _env_flag(SKIP_FITNESS_CHECKS_ENV):
+        log.info(f"Fitness checks disabled via {SKIP_FITNESS_CHECKS_ENV}, skipping.")
         return
 
     # Defer GPU check auto-registration until fitness checks are about to run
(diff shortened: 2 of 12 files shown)


runpod/serverless/modules/rp_fitness.py
```diff
@@ -294,3 +390,58 @@ async def run_fitness_checks() -> None:
...
+    are disabled or deferred, inside a running event loop, and in child
+    processes (multiprocessing 'spawn' re-imports this module; the worker marks
+    itself done via env so children skip). Ordinary exceptions from running the
+    checks are logged and swallowed: a failure to run the checks must not stop
+    a worker from booting. A failing check still force-exits, which is the point.
+    """
+    if _env_flag(SKIP_FITNESS_CHECKS_ENV) or _env_flag(DEFER_FITNESS_CHECKS_ENV):
+        return
+
+    if not os.environ.get("RUNPOD_WEBHOOK_GET_JOB"):
```
deanq: The import-time gate keys only on `RUNPOD_WEBHOOK_GET_JOB`, but that misses the `_is_local` / `test_input` local-mode guard that previously protected these checks. A handler run with `rp_args.test_input` used to skip all fitness checks (local mode -> `run_worker` never called). Now `import runpod` (eager on `main`) runs `run_startup_fitness_checks()`, executes the built-in memory/disk/network/gpu checks, and any failure hard-kills the process via `os._exit(1)`.

Same hazard for any auxiliary CLI/process that imports `runpod` only for the API client while `RUNPOD_WEBHOOK_GET_JOB` is inherited in the environment -- it will now run worker fitness checks and can be killed.

Suggest gating the import-time pass on the same local-mode/test-input signal that `run_worker` uses, so local and non-worker imports stay exempt.

runpod/serverless/__init__.py
```diff
@@ -29,6 +29,10 @@
 
 log = RunPodLogger()
 
+# Check the environment here rather than in start(), which a handler module
+# only reaches after loading its model. No-op outside a real worker.
+run_startup_fitness_checks()
```
deanq: The whole fail-fast benefit assumes `import runpod` eagerly imports `runpod.serverless`. That holds on this PR's declared base (`main`), but the repo's active `feat/apps-sdk` line lazy-loads `serverless` via PEP 562 `__getattr__`. There, a typical handler -- `import runpod` -> load model -> `runpod.serverless.start(...)` -- won't trigger `serverless/__init__` until the `start()` line, i.e. *after* the multi-minute model load.

So on the apps-sdk line the checks fire no earlier than before, silently regressing, while the README/docs added in this PR assert "built-ins at import" / "any `import runpod` triggers it."

Which branch does this actually merge into? If it's the lazy-import line, either the docs need correcting or the trigger needs an explicit hook that doesn't depend on eager submodule import.

runpod/serverless/modules/rp_fitness.py
```diff
@@ -294,3 +390,58 @@ async def run_fitness_checks() -> None:
...
+    # post-import changes (set in the handler, too late to apply).
+    _config_snapshot.update({v: os.environ.get(v) for v in _CONFIG_ENV_VARS})
+
+    try:
+        # Own loop rather than asyncio.run: run() resets the thread's loop
+        # policy state, after which asyncio.get_event_loop() in handler code
+        # raises RuntimeError on Python 3.10+.
+        loop = asyncio.new_event_loop()
+        try:
+            loop.run_until_complete(run_fitness_checks(include_deferred=False))
```
deanq: Running the network check at import time means a transient failure hard-exits the process mid-`import runpod` (`_terminate_unhealthy` -> `os._exit(1)`). On a cold worker container whose network stack isn't up yet when the handler module is first imported, this turns a recoverable warm-up delay into a boot crash-loop.

Previously this ran in `run_worker` after `start()`, giving the container time to become ready. Consider keeping network (and other environment-readiness) checks on the post-`start()` path, or adding a bounded retry before terminating.

runpod/serverless/modules/rp_fitness.py
```diff
@@ -237,22 +314,40 @@ async def run_fitness_checks() -> None:
         A failing check terminates the process via os._exit(1); this function
         does not return in that case and does not raise SystemExit.
     """
+    if _env_flag(SKIP_FITNESS_CHECKS_ENV):
+        log.info(f"Fitness checks disabled via {SKIP_FITNESS_CHECKS_ENV}, skipping.")
+        return
+
+    if _config_snapshot:
+        _warn_late_config()
```
deanq: `run_startup_fitness_checks` populates `_config_snapshot` microseconds before calling `run_fitness_checks`, where this `if _config_snapshot: _warn_late_config()` re-compares all nine env vars against the values just captured -- guaranteed no change, no warning. It's pure overhead on the import path; the late-config warning is only meaningful on the later `run_worker` pass. Consider skipping `_warn_late_config()` when invoked from the import-time pass.

deanq (commented):
Review from /code-review (correctness + cleanup pass). Four findings, most centered on moving os._exit(1)-capable checks to import time. Lines re-anchored to the current diff.

justinwlin:
bugbot run

justinwlin:
## Update: lock file replaced with an inherited env marker, realtime hook removed

Summary of what changed since the last round of review, and how the four review points map to the current code.

### What the PR does now, in one paragraph

`import runpod` in a Serverless worker runs the four hardware checks that are safe before the handler exists (RAM, disk, CUDA version, native GPU test). A failure exits the process right there, before any model loads. Network, torch CUDA init, GPU benchmark, and customer checks still run at `start()`. When the import pass passes, the process sets `RUNPOD_EARLY_FITNESS_CHECKS_DONE=1`, and every child it launches inherits that and skips the pass.

### Why the lock file is gone

The shared `/tmp` lock, procfs identity, and persisted failure state solved one problem: child processes that re-import the handler (vLLM spawn workers, `multiprocessing`, subprocesses) repeating the 30-second GPU probe. An inherited env var solves the same problem with no shared state, no lock waits, and no sticky failures. The realtime lifespan hook came out too, since realtime workers never ran fitness checks on `main` and that should be its own decision.

### Review points

| Point | Status | Where |
|---|---|---|
| Import gate misses local mode and `--test_input` | Fixed | `is_early_check_eligible` requires both worker env vars and skips `RUNPOD_TEST`, `--test_input`, `--rp_serve_api` |
| apps-sdk lazy-loads `serverless`, so the trigger never fires early | Fixed | Trigger moved to top of `runpod/__init__.py`; independent of `serverless` import. apps-sdk will need to carry the one-line call when main is merged in |
| Network check at import turns cold-boot network delay into a crash loop | Fixed | `_network_check` is `@defer_to_worker_start` and retries within one bounded budget |
| Late-config comparison is overhead on the import pass | Fixed | `_refresh_late_config` only runs on the worker-start pass |
| Registration failures escape the hard-exit path (Copilot) | Fixed | Registration is atomic; worker-start setup errors go through `_fail_worker` |

### Examples

**Healthy worker.** Logs from a real RTX 4090 Serverless worker on this branch:

```
Running 4 fitness check(s)...
GPU binary test passed: 1 GPU(s) healthy (CUDA 12.8)
Memory check passed: 184.90GB available (of 247.52GB total)
Disk space check passed: 19.78GB free (98.9% available)
CUDA version check passed: 12.8 (minimum: 11.8)
All fitness checks passed. (599.98ms)
IMPORT_DONE 3.82s marker=1
--- pretend model load ---
--- Starting Serverless Worker |  Version 1.12.1.dev23 ---
Running 3 fitness check(s)...
Network connectivity passed: Connected to api.runpod.ai:443
CUDA initialization passed: 1 device(s) initialized successfully
GPU compute benchmark passed: Matrix multiply completed in 108ms
```

Four checks inside the import, three at `start()`, nothing runs twice.

**Broken worker.** Same endpoint with `RUNPOD_MIN_MEMORY_GB=99999` to force a failure:

```
Running 4 fitness check(s)...
GPU binary test passed: 1 GPU(s) healthy (CUDA 12.8)
Fitness check failed: _memory_check | RuntimeError: Insufficient memory: 207.61GB available, 99999.0GB required
Worker is unhealthy, exiting.
```

Exit about one second into `import runpod`. The model-load line never prints. The platform marked the worker UNHEALTHY within a minute.

**Child process.** A handler that spawns a `multiprocessing` child, which re-imports the handler and therefore `runpod`:

```
Running 4 fitness check(s)...        <- parent, once
All fitness checks passed. (416.35ms)
PARENT import 2.29s marker=1
PARENT import 1.59s marker=1         <- child re-import, no checks
('child sees marker', '1')
```

**Opt-outs.** `RUNPOD_DEFER_FITNESS_CHECKS=true` restores the old worker-start timing. `RUNPOD_SKIP_FITNESS_CHECKS=true` disables all checks. Local `--test_input` runs and `RUNPOD_TEST` are exempt automatically.

### Validation

- 694 tests pass locally, 93% coverage, CI green on 3.10 through 3.14.
- Live on a Pod (RTX 4090): healthy import, forced failure, spawn child, and each exemption flag.
- Live on a Serverless endpoint (ADA_24): healthy cold start with a completed job, then a forced-failure release observed crash-looping at import.

justinwlin:
@copilot review

justinwlin:
```
  Process starts
      |
      v
  import runpod
      |
      +-- Realtime, local/test run, or checks disabled?
      |       |
      |       +-- Yes --> Skip early checks
      |
      +-- Checks deferred until worker start?
      |       |
      |       +-- Yes --> Wait until worker start
      |
      +-- Otherwise, run early checks:
              RAM → disk → CUDA version → GPU hardware
                  |
                  +-- Failure --> Exit unhealthy
                  |
                  +-- Pass --> Mark early checks done
                                 |
                                 v
                          Load the model
                                 |
                                 v
                          Start the worker
                                 |
                                 +-- Recheck disk space
                                 +-- Check network, CUDA initialization,
                                     GPU compute, and custom checks
                                        |
                                        +-- Failure --> Exit unhealthy
                                        |
                                        +-- Pass --> Accept jobs

  Child processes inherit the “early checks done” marker, so they skip the
  early checks when they import runpod again.
```

daveseddon-runpod (commented):
Nice work — the safety reasoning behind this change is the hard part, and it shows.

**👍 What's good**
- Safe/unsafe split is right: `@defer_to_worker_start` on `_cuda_init_check`/`_benchmark_check` keeps a CUDA context out of the pre-fork import path (vLLM/DeepSpeed), with docstrings explaining *why*. Network check deferred too.
- `os._exit` vs `sys.exit` documented with the real reason (live non-daemon threads blocking cooperative shutdown).
- Child marker `RUNPOD_EARLY_FITNESS_CHECKS_DONE` replacing the lock-file/procfs machinery — clean simplification; "marker only ever means a parent passed" is a tidy invariant.
- Security: `log.secret()` now fully `[REDACTED]` (won't `str()` the value) — closes the CodeQL clear-text finding and the old length side-channel.
- `sys.modules` aliases keep every old import path working.
- Latch-on-success registration (`_register_builtins` rolls back so malformed config re-raises loudly, not silently) — good catch, and it's tested.
- Docs (README, ARCHITECTURE, `worker_fitness_checks.md`) updated in step.

**🔴 Discuss before merge**
- **`import runpod` can now `os._exit(1)` any importer in a worker container** — not just the worker (sidecar scripts, spawn re-imports, a debug shell). Can we (a) call this out in `_startup.py`'s docstring / README gotchas, and (b) confirm the gate (`RUNPOD_ENDPOINT_ID` + `RUNPOD_WEBHOOK_GET_JOB` minus test flags) is tight enough that no standard-image tooling trips it? Biggest behavioral change here.
- **`_report_unhealthy` dropped the shared HTTP client + `WORKER_ID`** for raw `requests.Session()` + `os.environ["RUNPOD_POD_ID"]`. Intended? It loses any TLS/proxy/UA/retry defaults the shared client carries, and if `$RUNPOD_POD_ID` is in the URL but unset, it now sends nothing (vs a `WORKER_ID` fallback). Best-effort, so low impact — but lazily reusing the shared client would be safer.
- **Type the check metadata** (ties to modern types): flags live as dynamic attrs (`_runpod_defer_to_worker_start`, `_runpod_recheck_at_worker_start`, `_runpod_builtin`); `mypy --check-untyped-defs` flags every read. A small `@dataclass`/`Protocol` wrapping callable + flags makes the states typed and discoverable and kills the "typo in a `getattr` default silently disables a check" class of bug.

**🟡 Medium**
- **`nvidia-smi` probed 3+ times on the startup path** (`cuda.is_available()`, `gpu.auto_register_gpu_check()`, `system._get_cuda_version()`), each with a multi-second timeout. Cache detection (e.g. `lru_cache`) — a DRY win that also speeds the startup this PR targets.
- **`configure()` reassigns 5–7 module globals via `global` (`PLW0603`).** With the above, a frozen config object built from env once would be more testable and type-checked.
- **`binary_error` typing (`gpu.py`):** inferred `FileNotFoundError | None`, later reassigned `Exception`. Annotate `binary_error: Exception | None = None`.

**🟢 Low / nits**
- `_logger.py` uses `Optional[str]`; project is `>=3.10`, so `str | None` is idiomatic (`ruff UP045`) and drops the `typing` import.
- `ruff PLR2004`/`PLR0912` in `_get_cuda_version` — a named constant / small split reads better.
- No `CHANGELOG.md` entry despite new env vars + import-time behavior — worth one (confirm repo convention).
- Remaining `ruff` hits (`PLC0415`, `S110`, `S607`, `PERF203`) are intentional/correct here — noting them only to show they were reviewed, not missed.

**🔒 Security**
- Redaction improvement is solid; corner cases (empty, 1-char, `None`, `__str__`-raises) are tested.
- `_network_probe_target` avoids logging URL creds and fails safe on an unparseable URL.
- `_report_unhealthy` truncates `reason` to 256 chars to the same host as the heartbeat — low risk; note the exception message could carry paths.
- No new privilege: `RUNPOD_SKIP_FITNESS_CHECKS` is env-gated like existing controls.

**✅ Tests**
- Strong coverage: eligibility, realtime exclusion, deferred-vs-early, marker inheritance (real `multiprocessing` spawn), disk-recheck-at-start, network retry/bounded-close, malformed-config re-raise, redaction corner cases.
- `test_secret_redacts_short_empty_and_object_values` is effectively table-driven. Two asks to match our convention: (a) give each row an explicit `description`/`expected` so failures name the case; (b) fold the separate `_network_check`/URL-fallback tests into one `@pytest.mark.parametrize` table.

Overall: strong PR. Only real gates are the two 🔴 items; the rest are improvements.

<details>
<summary>Static analysis run (commands + results)</summary>

```console
$ ruff check --select E,F,W,B,UP,SIM,C4,ASYNC,S,PERF,RUF,PL --line-length 100 runpod/
# 36 findings, by rule:
  12  PLC0415  import should be at top-level         (intentional: deferred to break import cycle)
   7  UP045    use `X | None` for annotations        (_logger.py — modernize, drop `Optional`)
   7  PLW0603  `global` update in configure()        (system.py / gpu.py — see Medium)
   3  S607     partial executable path ("nvidia-smi")(accepted: PATH lookup of a system tool)
   2  S110     try/except/pass                       (accepted: best-effort before os._exit)
   2  PERF203  try/except within loop                (nvcc/nvidia-smi fallback — minor)
   1  PLW2901  loop var `line` overwritten           (gpu.py:82 — harmless)
   1  PLR2004  magic value `2`                       (system.py:243)
   1  PLR0912  too many branches (13>12)             (_get_cuda_version)

$ mypy --check-untyped-defs --explicit-package-bases runpod/_health/ runpod/_logger.py runpod/_startup.py
runpod/_health/fitness.py:121 error: "Callable[..., Any]" has no attribute "_runpod_defer_to_worker_start"
runpod/_health/fitness.py:131 error: "Callable[..., Any]" has no attribute "_runpod_recheck_at_worker_start"
runpod/_health/fitness.py:261 error: "Callable[..., Any]" has no attribute "_runpod_builtin"
runpod/_health/fitness.py:289 error: "Callable[..., Any]" has no attribute "_runpod_builtin"
runpod/_health/gpu.py:88/90/106/110 error: loose dict typing in _parse_gpu_test_output ("object" ...)
runpod/_health/gpu.py:259/263 error: binary_error type too narrow (FileNotFoundError | None <- Exception)
```

The `mypy` check-metadata errors are the concrete backing for 🔴 #3; `UP045` backs Low nit #1; `PLW0603` backs Medium; the `binary_error` errors back Medium. The remaining `ruff` hits are the intentional-and-accepted set noted under Low.
</details>

deanq (approved):
Thanks for the rework, all four points are addressed. The early-check gate now excludes realtime workers and test_input/serve-api runs and keys on both `RUNPOD_ENDPOINT_ID` and `RUNPOD_WEBHOOK_GET_JOB`, and moving `run_import_checks()` into `runpod/__init__.py` ahead of the serverless import makes it fire independent of eager vs lazy submodule loading, so the apps-sdk lazy path is covered. The network probe is deferred off the import path via `@defer_to_worker_start` and now retries within a bounded timeout before failing, which removes the cold-container crash-loop. And `_refresh_late_config` only runs on the deferred worker-start pass now, so the import pass no longer does the guaranteed-noop comparison. The env-marker coordination reads cleaner than the old lock file and keeps the failure-never-marks invariant. LGTM.

PR #584, opened 2026-09-02
feat: migrate api wrapper to rest v2

KAJdev:
Routes the existing API helpers through REST API v2 while preserving the public function surface and list unwrapping behavior. GraphQL is scoped to account profile reads and registry credential updates, since those operations are not exposed by REST API v2.

Linear: CON-1154


runpod/api/ctl_commands.py
```diff
@@ -124,151 +126,117 @@ def create_pod(
...
+        body["mounts"] = {
+            "network": [{"volumeId": network_volume_id, "path": mount_path}]
+        }
+    elif volume_in_gb:
+        body["mounts"] = {"persistent": {"size": volume_in_gb, "path": mount_path}}
+    elif template_id and gpu_type_id and volume_mount_path is not None:
+        template = run_rest_request(
+            "GET", f"/v2/templates/{_path_segment(template_id)}"
+        )
+        persistent = template["mounts"].get("persistent")
```
deanq: `template["mounts"].get("persistent")` will raise `KeyError` when the template has no persistent volume — the REST v2 template response omits the `mounts` key entirely in that case, rather than returning it empty. A `create_pod` call with `template_id` + `volume_mount_path` against such a template crashes here instead of proceeding.

Suggest `template.get("mounts", {}).get("persistent")`.

runpod/api/ctl_commands.py
```diff
@@ -124,151 +126,117 @@ def create_pod(
...
-            env,
-            template_id,
-            network_volume_id,
-            allowed_cuda_versions,
-            min_download,
-            min_upload,
-            instance_id,
-        )
+    unsupported = []
+    if support_public_ip is not True:
```
deanq: Rejecting `support_public_ip=False` with a `ValueError` is a regression: under the GraphQL API this was a valid way to provision a pod without a public IP. The `if support_public_ip is not True` guard now raises on it. Separately, `support_public_ip=True` is never forwarded to the request body, so public-IP behavior is left entirely to the REST default — worth confirming that default matches the old behavior, or dropping the parameter from the signature if it's genuinely unsupported.

runpod/api/ctl_commands.py
```diff
@@ -344,98 +297,82 @@ def create_endpoint(
...
-    >>> endpoint_id = runpod.update_endpoint_template("test", "template_id")
-    """
-    raw_response = run_graphql_query(
-        endpoint_mutations.update_endpoint_template_mutation(endpoint_id, template_id)
+) -> dict:
+    """Create a queue-based serverless endpoint; locations are datacenter IDs."""
+    scaler_type = {
+        "QUEUE_DELAY": "QUEUE_DELAY",
+        "REQUEST_COUNT": "REQUEST_COUNT",
+        "WORKER_COUNT": "REQUEST_COUNT",
```
deanq: Mapping `WORKER_COUNT` → `REQUEST_COUNT` silently changes autoscaling semantics: a caller who set `scaler_type='WORKER_COUNT'` with `scaler_value=N` meant N workers, but this now treats N as a request-count target. That's a behavior change with no warning or error — if REST v2 genuinely dropped worker-count scaling, this should raise (like the unsupported-options guard above) rather than silently remap to a different policy.

runpod/api/ctl_commands.py
```diff
@@ -344,98 +297,82 @@ def create_endpoint(
...
+        "count": gpu_count,
+    }
+    if excluded_types:
+        gpu["excludedTypes"] = excluded_types
+    if allowed_cuda_versions is not None:
+        gpu["allowedCudaVersions"] = _split_values(allowed_cuda_versions)
+
+    workers = {"min": workers_min, "max": workers_max}
+    if scaler_type == "QUEUE_DELAY":
+        workers["idleTimeout"] = idle_timeout
```
deanq: `idle_timeout` only applies to the `QUEUE_DELAY` scaler type, so gating `workers["idleTimeout"]` on that branch is correct. The one gap: when a caller passes `idle_timeout` alongside a different scaler type, it's silently dropped. Worth a warning or a `ValueError` there so the caller knows the value was ignored rather than applied.

runpod/cli/groups/project/functions.py
```diff
@@ -60,10 +60,7 @@ def _launch_dev_pod():
     sys.stdout.flush()
 
     # Wait for the pod to come online
-    while (
-        new_pod.get("desiredStatus", None) != "RUNNING"
-        or new_pod.get("runtime") is None
-    ):
+    while new_pod.get("status") != "RUNNING" or new_pod.get("runtime") is None:
         new_pod = get_pod(new_pod["id"])
```
deanq: Now that `get_pod` returns `None` on a 404 instead of raising, this poll loop can crash: if a just-created pod transiently 404s, `new_pod = get_pod(...)` becomes `None` and the next iteration's `new_pod.get("status")` raises `AttributeError: 'NoneType' object has no attribute 'get'`. Guard for `None` (treat it as not-ready-yet and keep polling).

runpod/api/ctl_commands.py
```diff
@@ -1,113 +1,115 @@
...
+        json={"keys": keys},
     )
-    return raw_response["data"]["pod"]
+    return get_user(api_key=api_key)
+
+
+def get_gpus(api_key: Optional[str] = None) -> list[dict]:
+    """Get all GPU types."""
+    response = run_rest_request("GET", "/v2/catalog/gpus", api_key=api_key)
+    return response["gpus"]
```
deanq, line 69: `run_rest_request` returns `None` on a 204/empty-body response, so `return response["gpus"]` raises `TypeError: 'NoneType' object is not subscriptable` if the API ever returns no content. Same pattern at `response["pods"]` (line 100) and `response["endpoints"]` (line 283). Default to an empty result (`(response or {}).get("gpus", [])`) unless a missing body should genuinely be an error.

runpod/api/ctl_commands.py
```diff
@@ -124,151 +126,117 @@ def create_pod(
...
-
-    if gpu_type_id is not None:
-        cleaned_response = raw_response["data"]["podFindAndDeployOnDemand"]
+    if network_volume_id:
+        body["mounts"] = {
+            "network": [{"volumeId": network_volume_id, "path": mount_path}]
+        }
+    elif volume_in_gb:
+        body["mounts"] = {"persistent": {"size": volume_in_gb, "path": mount_path}}
+    elif template_id and gpu_type_id and volume_mount_path is not None:
```
deanq, line 192: The `elif template_id and gpu_type_id and volume_mount_path is not None` branch only resolves the template-derived mount when all three are set — notably requiring `gpu_type_id`. A CPU pod (no GPU type) or a call omitting `volume_mount_path` silently skips template mount resolution. Consider resolving the template's storage uniformly and applying path/size overrides on top, rather than gating the whole resolution on this multi-condition special case.

runpod/api/rest.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _raise_for_error(
+    response: requests.Response, method: str, path: str
+) -> None:
+    if response.status_code == HTTP_STATUS_UNAUTHORIZED:
+        raise error.AuthenticationError(
+            "Unauthorized request, please check your API key."
+        )
+
+    if response.status_code < 400:
```
deanq: rest.py already defines `HTTP_STATUS_UNAUTHORIZED` / `HTTP_STATUS_NO_CONTENT`, but this uses a bare `< 400` and ctl_commands.py uses bare `== 404` (lines 89/110). For consistency, pull the not-found and error-threshold codes into named constants too.

deanq (commented):
A few correctness issues from the GraphQL to REST v2 migration, plus a couple of smaller notes. The silent behavior changes in create_endpoint and the KeyError in create_pod are the ones I'd want addressed before merge.

runpod/api/ctl_commands.py
deanq: Still open after the latest push — `get_gpus`/`get_pods`/`get_endpoints` all still do `return response["<key>"]` (lines 69/97/283). Low-severity since these list endpoints likely never return a 204, so I'm fine leaving it if that's deliberate. If you want a cheap guard, `(response or {}).get("gpus", [])` gives callers an empty list instead of a `TypeError` should the API ever change.

runpod/api/ctl_commands.py
deanq: Still unchanged (line 192). This one's a design nit, not a bug — happy to leave it for a follow-up. Flagging only so it's a conscious decision: a CPU pod or a call without `volume_mount_path` silently skips template mount resolution, because the branch requires `gpu_type_id`.

runpod/api/mutations/container_register_auth.py
```diff
@@ -1,51 +1,18 @@
...
+) -> str:
+    """Build the registry credential update mutation."""
+    input_dict = {
+        "id": registry_auth_id,
+        "username": username,
+        "password": password,
+    }
+    input_str = ", ".join(
+        f'{key}: "{value}"' for key, value in input_dict.items()
+    )
```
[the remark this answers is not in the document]
deanq: Seconding this — it's a real injection vector. `update_container_registry_auth` still routes through `run_graphql_query` with these values interpolated into the mutation string, so a credential containing a quote, backslash, or newline breaks the query (or worse). The create path moved to REST in this PR; this update path should either migrate to REST too or use GraphQL variables instead of string interpolation.

runpod/api/ctl_commands.py
KAJdev: CPU pods do not support local persistent volumes. It wont skip the template mount, the template ID is sent above always, this branch is only for catching the specific case when someone provides a template id, and specified a mount path override for a volume but not a size override, so that has to be fetched so its a valid override.

runpod/api/ctl_commands.py
KAJdev: probably going to keep these as is, an empty list has meaning, if an empty list is not sent then that's a malformed response and should be treated as such.

runpod/api/mutations/container_register_auth.py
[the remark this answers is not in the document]
KAJdev: this is already fixed https://github.com/runpod/runpod-python/blob/df782193994b47195d28d28f84f85ec6ca91a00e/runpod/api/mutations/container_register_auth.py

deanq (approved):
LGTM. The REST v2 migration reads cleanly and my earlier concerns are all addressed:

- Registry credential injection is closed — `json.dumps` on the values produces valid escaped GraphQL string literals. Migrating this update path to REST/variables would be a nice follow-up but the escaping is correct as-is.
- The empty-body and template-mount points are reasonable as deliberate decisions given your explanations (empty list is meaningful; CPU pods don't take local persistent volumes and template_id is always forwarded).

Nothing blocking from me.

PR #587, opened 2026-09-03
fix: stream errors to client before breaking generator loop

### LINKED ISSUE #397: Breaking change in error handling in Serverless API with handler using generator
**Describe the bug**
After upgrading to the 1.7.7 version of the Runpod python package, I'm facing a breaking change in the way errors are handled when using streaming in a serverless endpoint and my handler is using an async generator: errors are no longer sent in the streaming output, I have to call the status endpoint to get the error details.

This seems to be related to changes in the PR https://github.com/runpod/runpod-python/pull/384. While I understand the purpose of this PR, which is to return a FAILED status instead of a COMPLETED one, which is great, it seems to introduce two side effects in my project:
1- Errors are no longer sent to my streaming output (because the `stream_result` function is not called anymore),
2- I was sending information along the `error` field that I can no longer retrieve (because only the `error` field is kept in the `stream_output`), in particular the `refresh_worker` field. However, when I look at the code, it seems that the `refresh_worker` field is not handled in generators (but this is another issue).

**To Reproduce**
1. Start from example described here: https://docs.runpod.io/serverless/workers/handlers/handler-async#writing-asynchronous-handlers
2. Replace the `yield output` line with:
```
yield {
  "error": "An error has occurred.",
  "context": "Some context.",
  "refresh_worker": True
}
```
3. Create a new job by using the `run` endpoint as described [here](https://docs.runpod.io/serverless/endpoints/job-operations#asynchronous-endpoints).
4. Use the `stream` endpoint as described [here](https://docs.runpod.io/serverless/endpoints/job-operations#stream-results) to get the streaming output and check that it is empty but with a FAILED status in runpod version 1.7.7 whereas previous versions contains the yielded output but with a COMPLETED status.
5. Use the `status` endpoint as described [here](https://docs.runpod.io/serverless/endpoints/job-operations#check-job-status) to get the previously yielded `error` (but without the `context` field).

**Expected behavior**
I'm expecting to get the error in the streaming result to avoid a new call to the `status` API endpoint, as well as being able to add context to the error output (although I can send it using the error field).

Otherwise, maybe add some documentation explaining how to retrieve errors when using streaming and handlers with a generator? Or maybe there is another way of getting my error information?

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

Swigler:
## Summary

- **Stream errors are now delivered to clients** — `stream_result()` is called with the error before breaking out of the generator loop
- **Extra fields are preserved** — fields like `context` and `refresh_worker` survive error extraction instead of being stripped

Fixes #397

## What was wrong

Since v1.7.7 (PR #384), when a generator handler yields an error, the code in `handle_job()` breaks out of the streaming loop *before* calling `stream_result()`:

```python
if stream_output.get("error"):
    job_result = stream_output
    break                          # ← exits loop

await stream_result(...)           # ← never reached for errors
```

Clients listening on `/stream` receive normal chunks fine, but when an error occurs the stream goes silent. The error is saved internally (visible via `/status`), but never delivered through the stream.

Additionally, the error extraction replaced the entire dict:
```python
stream_output = {"error": str(stream_output["output"]["error"])}
```
This discarded all other fields (`context`, `refresh_worker`, etc.).

PR #384's intent (marking errored jobs as FAILED instead of COMPLETED) is preserved — `send_result()` still receives `job_result` with the error and sets the correct status.

## What changed

**`rp_job.py` — `handle_job()` generator loop:**
- Call `await stream_result(session, stream_output, job)` before `break` so errors reach the stream
- Preserve extra fields from the error dict using dict unpacking instead of replacing the whole dict

**`test_job.py` — 2 new tests:**
- Error is streamed before break (stream_result called with error payload)
- Nested error preserves extra fields (context, refresh_worker survive extraction)

## Test plan

- [x] All 31 tests in `test_job.py` pass (29 existing + 2 new)
- [x] New test verifies `stream_result` is called twice (once for normal chunk, once for error)
- [x] New test verifies extra fields (`context`, `refresh_worker`) are present in streamed error

### DIFF
--- a/runpod/serverless/modules/rp_job.py
+++ b/runpod/serverless/modules/rp_job.py
@@ -204,9 +204,18 @@ async def handle_job(session: ClientSession, config: Dict[str, Any], job) -> dic
 
             if isinstance(stream_output.get("output"), dict):
                 if stream_output["output"].get("error"):
-                    stream_output = {"error": str(stream_output["output"]["error"])}
+                    error_fields = {
+                        k: v
+                        for k, v in stream_output["output"].items()
+                        if k != "error"
+                    }
+                    stream_output = {
+                        "error": str(stream_output["output"]["error"]),
+                        **error_fields,
+                    }
 
             if stream_output.get("error"):
+                await stream_result(session, stream_output, job)
                 job_result = stream_output
                 break
 
--- a/tests/test_serverless/test_modules/test_job.py
+++ b/tests/test_serverless/test_modules/test_job.py
@@ -440,3 +440,71 @@ class TestRunJobGenerator(IsolatedAsyncioTestCase):
         assert mock_log.info.call_count == 1
         mock_log.info.assert_called_with("Finished running generator.", "123")
 
+
+class TestHandleJobStreamError(IsolatedAsyncioTestCase):
+    """Tests that generator errors are streamed to the client. (Fixes #397)"""
+
+    @staticmethod
+    async def _noop(*args, **kwargs):
+        pass
+
+    async def test_error_is_streamed_before_break(self):
+        """When a generator yields an error, stream_result must be called with it."""
+
+        def error_gen(job):
+            yield "chunk_1"
+            yield {"error": "GPU OOM"}
+
+        config = {"handler": error_gen}
+        job = {"id": "test-123"}
+        session = Mock()
+
+        stream_calls = []
+
+        async def capture_stream(*args, **kwargs):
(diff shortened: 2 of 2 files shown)


Swigler:
Ping — errors during streaming are swallowed because the generator breaks before sending. This streams the error to the client first. Ready for review.

